PR #8, opened 2021-01-27
Fix Magazine Scheduled Content query

B77Mills:
Only set section on query if it is going to be used or else it will result in 0 results because section will never match an empty object. :(

### DIFF
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -912,30 +912,21 @@ module.exports = {
       } = input;
 
       const since = new Date();
-      const idQuery = {
-        issue: issueId,
-      };
-      if (sectionId || includeSectionNames.length || excludeSectionNames.length) idQuery.section = {};
-      if (sectionId) idQuery.section.$eq = sectionId;
-      if (includeSectionNames.length) {
-        const includeSections = await loadMagazineSections({
-          basedb,
-          names: [...includeSectionNames],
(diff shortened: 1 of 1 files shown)


services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -914,8 +914,8 @@ module.exports = {
       const since = new Date();
       const idQuery = {
         issue: issueId,
-        section: {},
       };
+      if (sectionId || includeSectionNames.length || excludeSectionNames.length) idQuery.section = {};
```
solocommand: Just to consolidate a bit, give this a try:
```js
      const idQuery = { issue: issueId };
      if (sectionId || includeSectionNames.length || excludeSectionNames.length) {
        const [include, exclude] = await Promise.all([
          includeSectionNames.length ? loadMagazineSections({ basedb, names: includeSectionNames }) : [],
          excludeSectionNames.length ? loadMagazineSections({ basedb, names: excludeSectionNames }) : [],
        ]);
        idQuery.section = {
          ...(sectionId && { $eq: sectionId }),
          ...(include.length && { $in: include.map((section) => section._id ) }),
          ...(exclude.length && { $in: exclude.map((section) => section._id ) }),
        };
      }
      
      const ids = await basedb.distinct('magazine.Schedule', 'content.$id', idQuery);
```

solocommand (changes requested):

solocommand (approved):

PR #28, opened 2021-02-26
Emit event when Leaders category companies are loaded

zarathustra323:
When a Leaders category is expanded, an event will fire that includes the category's section information and the companies listed with the category. This event will _not_ fire if no companies were found for the expanded section


zarathustra323:
[Add leader (company) "impression" tracking](https://trello.com/c/O8tVnZMP/6-add-leader-company-impression-tracking)

solocommand (approved):

PR #38, opened 2021-03-05
Create inquiry submission tracker

zarathustra323:
Pushes the inquiry payload to P1 events


zarathustra323:
[Create RMI form hook to submit lead data to P1 events](https://trello.com/c/N6igvUB7/18-create-rmi-form-hook-to-submit-lead-data-to-p1-events)

brandonbk (approved):

PR #39, opened 2021-03-09
Windows development updates

solocommand:
- [x] Changes all `0.0.0.0` hostname references to `localhost` for cross-platform compatibility.
- [x] Allows injection of `exposedHost` parameter in addition to `exposedPort`.

### DIFF
--- a/packages/export-framework/start-server.js
+++ b/packages/export-framework/start-server.js
@@ -9,9 +9,8 @@ const CoreConfig = require('./config/core');
 const CustomConfig = require('./config/custom');
 
 if (!process.env.LIVERELOAD_PORT) process.env.LIVERELOAD_PORT = 5010;
-if (!process.env.LIVERELOAD_HOST) {
-  process.env.LIVERELOAD_HOST = 'localhost';
-}
+if (!process.env.LIVERELOAD_HOST) process.env.LIVERELOAD_HOST = 'localhost';
+
 const { env } = process;
 process.on('unhandledRejection', (e) => { throw e; });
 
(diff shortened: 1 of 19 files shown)


packages/export-framework/start-server.js
```diff
@@ -9,6 +9,8 @@ const CoreConfig = require('./config/core');
 const CustomConfig = require('./config/custom');
 
 if (!process.env.LIVERELOAD_PORT) process.env.LIVERELOAD_PORT = 5010;
+if (!process.env.LIVERELOAD_HOST) process.env.LIVERELOAD_HOST = 'localhost';
```
zarathustra323: Can we change all refs to the live reload host to use the exposed host env instead? Live reload should simply use the same host as the service.

zarathustra323 (changes requested):

packages/export-framework/start-server.js
```diff
@@ -19,6 +21,7 @@ module.exports = async ({
   coreConfig: incomingCoreConfig,
   port = env.PORT || 6008,
   exposedPort = env.EXPOSED_PORT || env.PORT || 6008,
+  exposedHost = env.EXPOSED_HOST || 'localhost',
```
zarathustra323: Since the env value is being set above there's no need to do the fallback to localhost now

packages/marko-newsletters/start-server.js
```diff
@@ -24,6 +24,7 @@ module.exports = async ({
   coreConfig,
   port = env.PORT || 5008,
   exposedPort = env.EXPOSED_PORT || env.PORT || 5008,
+  exposedHost = env.EXPOSED_HOST || 'localhost',
```
zarathustra323: Same here

packages/marko-web/start-server.js
```diff
@@ -23,6 +23,7 @@ module.exports = async ({
   helmetConfig,
   port = env.PORT || 4008,
   exposedPort = env.EXPOSED_PORT || env.PORT || 4008,
+  exposedHost = env.EXPOSED_HOST || 'localhost',
```
zarathustra323: And here

zarathustra323 (changes requested):
Can you also confirm that these changes still work in a Mac environment?

solocommand:
image
yes it still works fine

zarathustra323 (approved):

PR #40, opened 2021-03-09
Windows file watching

solocommand:
- [x] new: Provides support for alternate file watching (specifically for Docker on Windows)
- [x] bugfix: Only attempt to execute callback in cli-utils if it is a function
- [x] new: allow specifying ignore paths for gulp via fn input
- [x] bugfix: do not attempt to lint `node_modules` when booting the newsletter cli
- [x] cleanup: Remove deprecated `eslint.results` calls

Consuming applications can now specify two additional environment variables to be picked up by the various CLIs:
- `GULP_POLLING_ENABLED` - If set, the file watching method will use polling instead of the default fs.watch.
- `GULP_POLLING_INTERVAL` - (Default 1000ms) Sets the interval that watched files will be polled for changes (via modification dates)

image

image

### DIFF
--- a/packages/export-cli/src/gulp/lint.js
+++ b/packages/export-cli/src/gulp/lint.js
@@ -4,15 +4,11 @@ const pump = require('pump');
 const { src } = require('gulp');
 const completeTask = require('@parameter1/base-cms-cli-utils/task-callback');
 
-module.exports = (cwd, options) => (cb) => {
+module.exports = (cwd, options, ignore = ['**/node_modules/**/*']) => (cb) => {
   pump([
-    src(['**/*.js'], { cwd }),
+    src(['**/*.js'], { cwd, ignore }),
     cache('basecms-exports-lint-js'),
     eslint(options),
-    eslint.results((results, lintCb) => {
-      lintCb();
-      cb();
-    }),
     eslint.format(),
   ], e => completeTask(e, cb));
 };
--- a/packages/export-cli/src/gulp/serve.js
+++ b/packages/export-cli/src/gulp/serve.js
@@ -9,6 +9,7 @@ const lint = require('./lint');
 const server = require('./server');
 
 module.exports = (cwd, serverFile) => () => {
+  if (process.env.GULP_POLLING_ENABLED) log('Falling back to polling to watch files!');
   const watcher = watch(
     ['**/*.js'],
     {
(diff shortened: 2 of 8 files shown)


packages/export-cli/src/gulp/lint.js
```diff
@@ -4,15 +4,11 @@ const pump = require('pump');
...
 const completeTask = require('@parameter1/base-cms-cli-utils/task-callback');
 
-module.exports = (cwd, options) => (cb) => {
+module.exports = (cwd, options, ignore = ['**/node_modules/**/*']) => (cb) => {
   pump([
-    src(['**/*.js'], { cwd }),
+    src(['**/*.js'], { cwd, ignore }),
     cache('basecms-exports-lint-js'),
     eslint(options),
-    eslint.results((results, lintCb) => {
```
zarathustra323, line 12: I honestly don't remember what the results callback (`lintCb`) was being invoked for... that being said, can you better explain why this call was removed?

packages/export-cli/src/gulp/serve.js
```diff
@@ -9,13 +9,21 @@ const lint = require('./lint');
...
+  if (process.env.GULP_POLLING_ENABLED) log('Falling back to polling to watch files!');
   const watcher = watch(
     ['**/*.js'],
     {
       cwd,
       queue: false,
       ignoreInitial: false,
       ignored: ['node_modules'],
+      ...(process.env.GULP_POLLING_ENABLED && {
+        // Use polling on windows https://forums.docker.com/t/file-system-watch-does-not-work-with-mounted-volumes/12038/16
```
zarathustra323, line 21: It looks like polling is going to cause significant resource issues. Have you explored the option presented in [this](https://forums.docker.com/t/file-system-watch-does-not-work-with-mounted-volumes/12038/21) comment (from the same thread)?

zarathustra323 (changes requested):

packages/export-cli/src/gulp/lint.js
solocommand: https://github.com/adametry/gulp-eslint/tree/v5.0.0#eslintresultsaction using the second form. This was added to allow calling the user-supplied `cb` function after the lint was complete, but it's unnecessary since `completeTask` below *also* calls the `cb` function once the stream is complete.

packages/export-cli/src/gulp/serve.js
solocommand: I didn't because I didn't want to introduce the extra dependencies it would require: a working Python environment, the `pip` package manager, and the `docker-windows-volume-watcher` utility itself.

zarathustra323 (approved):

PR #44, opened 2021-03-15
Remove empty tags before placing ads

brandonbk:
Example body copy: https://www.equipmentworld.com/alternative-power/article/15064224/test-article?preview-mode=true

### DIFF
--- a/packages/marko-web-gam/browser/inject-ads.vue
+++ b/packages/marko-web-gam/browser/inject-ads.vue
@@ -43,19 +43,12 @@ export default {
     const { childSelector } = this;
     const component = this;
     const $el = $(this.selector);
-
-    // Remove empty tags
-    $('p', $el).each(function removeEmptyTags() {
-      if ($(this).html().replace(/\s|&nbsp;/g, '').length === 0) {
-        $(this).remove();
-      }
-    });
-
     const $children = $(`> ${childSelector}`, $el);
 
     $children.each(function injectAds(index) {
       const $child = $(this);
-      const { length } = $child.text();
+      const { length } = $child.text().replace(/\s|&nbsp;/g, '');
+      if (length === 0) return;
       const $nextChild = $(this).next(childSelector);
 
       component.targetLengths.forEach((targetLength) => {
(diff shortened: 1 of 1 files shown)


brandonbk:
@zarathustra323 - penny for your thoughts here.  It feels wrong to approach this way, but I'm also not sure if "correcting" at the API layer or on save within Base is right either.

zarathustra323:
Instead of removing the element (since it will mutate the DOM) how about just skipping it - thus removing the empty tag from
consideration.  

Also, the empty selector will only literally match `<p></p>`. A paragraph with a blank space will not match nor will a paragraph with an empty child element. So perhaps you should also check the trimmed value of the jQuery `.text()` method.

zarathustra323 (changes requested):
See my previous comment :)

brandonbk:
[Filter empty tags in article body](https://trello.com/c/eTemzFmK/44-filter-empty-tags-in-article-body)

brandonbk:
@zarathustra323 - I ran several tests with this version, including multiple spaces, `nbsp;`, a combination of both, etc.

Seems to be working as intended.

Screen Shot 2021-04-11 at 8 48 03 PM

packages/marko-web-gam/browser/inject-ads.vue
```diff
@@ -43,6 +43,14 @@ export default {
     const { childSelector } = this;
     const component = this;
     const $el = $(this.selector);
+
+    // Remove empty tags
+    $('p', $el).each(function removeEmptyTags() {
+      if ($(this).html().replace(/\s|&nbsp;/g, '').length === 0) {
```
zarathustra323: Wouldn't `$(this).text().trim().length` be better? Also, it will be considerably safer to not remove the element from the active DOM (which this will do) and instead skip the element.

I believe you could simply add a `trim` to line 58, and then add a return statement on the next line if `length === 0`

zarathustra323 (changes requested):

packages/marko-web-gam/browser/inject-ads.vue
zarathustra323: This will also account for situations such as `<p><span></span></p>`

zarathustra323: Actually, the return statement isn't necessary: simply trimming the `.text()` value should suffice, as the character count will not be incremented (since the length will be `0`), thereby effectively "skipping" the empty element

PR #45, opened 2021-03-18
Restore eslint callbacks

solocommand:
Evidently I was wrong in #40 -- while the `completeTask` CLI utility will execute the user-supplied callback regardless of error state, the `pump` utility *only* calls the supplied callback (`completeTask`) if the stream encountered an error: https://github.com/mafintosh/pump/blob/master/index.js#L75

Reverts change to `eslint.results` calls to restore the callback.


zarathustra323 (approved):
So I guess I was right for asking about this :)

PR #60, opened 2021-03-30
NativeX Story query support

solocommand:
Adds support for querying for a NativeX story and rendering a template. By default, this simply renders the story image, title, and body. The template and query fragment can be overridden by passing them as arguments to the router:

```js
const NativeXStoryRouter = require('@parameter1/base-cms-marko-web-native-x/routes/story');
const template = require('../templates/native-x/story');
const queryFragment = require('../graphql/fragments/native-x-story');

module.exports = (app) => {
  const config = getAsObject(app, 'locals.nativeX');
  app.get('/sponsored/:section/:slug/:id', NativeXStoryRouter({ config, template, queryFragment }));
};
```


solocommand:
[Embed Native-X stories within the site](https://trello.com/c/rHbOMWJt/48-embed-native-x-stories-within-the-site)

packages/marko-web-native-x/utils/convert-story-to-content.js
```diff
@@ -0,0 +1,46 @@
...
+  return {
+    id: story.id,
+    name: story.title,
+    shortName: story.title,
+    body: story.body,
+    typeTitled: 'Text Ad',
+    type: 'text-ad',
+    teaser: story.teaser,
+    published: startDate,
+    canonicalPath: story.url, // @deprecated. Use siteContext.path instead
```
zarathustra323: Kill this since it's deprecated - no reason to keep it around

packages/marko-web-native-x/utils/create-client.js
```diff
@@ -0,0 +1,14 @@
+const fetch = require('node-fetch');
```
zarathustra323: Move this to an `apollo` folder so it's more clear what kind of client is being created

packages/marko-web-native-x/templates/story.marko
```diff
@@ -0,0 +1,39 @@
+import { buildImgixUrl } from "@parameter1/base-cms-image";
```
zarathustra323: Remove this - the site should implement the template - there's no need for a basic template as it will never be used by an implementing website

packages/marko-web-native-x/routes/story.js
```diff
@@ -0,0 +1,25 @@
...
+const defaultTemplate = require('../templates/story');
+const defaultFragment = require('../graphql/fragments/story');
+const buildQuery = require('../graphql/queries/story');
+
+module.exports = ({
+  config,
+  template = defaultTemplate,
+  queryFragment = defaultFragment,
+} = {}) => asyncRoute(async (req, res, next) => {
+  if (config.isEnabled()) {
```
zarathustra323: No need for an enable check... if NX isn't enabled, the site implementor shouldn't include the middleware to begin with

packages/marko-web-native-x/routes/story.js
```diff
@@ -0,0 +1,25 @@
+const { asyncRoute } = require('@parameter1/base-cms-utils');
```
zarathustra323: This should be moved to a `middleware` folder, not `routes`

packages/marko-web-native-x/routes/story.js
```diff
@@ -0,0 +1,25 @@
...
+const { getAsObject } = require('@parameter1/base-cms-object-path');
+const createClient = require('../utils/create-client');
+const defaultTemplate = require('../templates/story');
+const defaultFragment = require('../graphql/fragments/story');
+const buildQuery = require('../graphql/queries/story');
+
+module.exports = ({
+  config,
+  template = defaultTemplate,
```
zarathustra323: Require a `template` to be passed - do not use a default

packages/marko-web-native-x/utils/convert-story-to-content.js
```diff
@@ -0,0 +1,46 @@
...
+  const advertiser = getAsObject(story, 'advertiser');
+  const contentImage = getAsObject(story, 'primaryImage');
+  const advertiserImage = getAsObject(story, 'advertiser.logo');
+  const startDate = get(story, 'publishedAt');
+  return {
+    id: story.id,
+    name: story.title,
+    shortName: story.title,
+    body: story.body,
+    typeTitled: 'Text Ad',
```
zarathustra323: The default type should be an Article, not a Text Ad...

Can you double-check all of these to make sure they work with a Story, since this was copied-pasted from the campaign builder? :)

zarathustra323 (changes requested):

packages/marko-web-native-x/routes/story.js
zarathustra323: Rename the file to `with-story.js` once moved to the middleware folder.  When imported by the site, the usage should be this:

```js
const withNativeXStory = require('@parameter1/base-cms-marko-web-native-x/middleware/with-story.js');
const template = require('../templates/my-site-template');

app.use('/some-path', withNativeXStory({ template }));
```

zarathustra323 (changes requested):

packages/marko-web-native-x/templates/story.marko
zarathustra323: If anything, move this to the README as an example/documatation

zarathustra323 (approved):

PR #61, opened 2021-03-31
Add GTM container id to Nx middleware

solocommand:

PR #62, opened 2021-03-31
NativeX story GTM/event tracking

solocommand:
Supercedes #61 -- provides reusable components to track NativeX story events

- [x] GTM initialization component
- [x] Track page view
- [x] Track end of content
- [x] Track social share (bridge to package events
- [x] Track outbound links


packages/marko-web-native-x/components/gtm-init.marko
```diff
@@ -0,0 +1,27 @@
+import defaultValue from "@parameter1/base-cms-marko-core/utils/default-value";
+import { get, getAsObject } from "@parameter1/base-cms-object-path";
+
+$ const { req } = out.global;
+$ const story = getAsObject(input, "story");
+$ const on = defaultValue(input.on, "load");
+$ const requestFrame = defaultValue(input.requestFrame, true);
+$ const targetTag = defaultValue(input.targetTag, "body");
+$ const push = {
```
zarathustra323: Now that i see that the push data needs the story, perhaps it would be better to move the push into its own track page view component. That way the user has a choice where to use it, and it keeps the init component cleaner

zarathustra323 (changes requested):

packages/marko-web-native-x/browser/track-end-of-content.vue
```diff
@@ -0,0 +1,44 @@
...
+      default: false,
+    },
+  },
+  created() {
+    this.observer = new IntersectionObserver((event) => {
+      if (event[0].isIntersecting) {
+        endOfContent(window);
+      }
+    });
+    const dl = window.dataLayerNativeX || [];
```
zarathustra323: This should detect the presence of the data layer. If it doesn't exist, nothing should happen... a new queue shouldn't be created

packages/marko-web-native-x/utils/gtm-events.js
```diff
@@ -0,0 +1,19 @@
+export const pageLoad = (window) => {
+  const dataLayer = window.dataLayerNativeX || [];
```
zarathustra323: All of these functions should detect the presence of the data layer and do nothing if it doesn't exist

zarathustra323 (changes requested):

packages/marko-web-native-x/browser/track-outbound-links.vue
```diff
@@ -0,0 +1,26 @@
...
+      type: String,
+      required: true,
+    },
+  },
+  created() {
+    const elements = document.querySelectorAll(`${this.container} a[href]`);
+    elements.forEach((element) => {
+      const url = element.getAttribute('href');
+      element.addEventListener('click', (e) => {
+        e.preventDefault();
```
zarathustra323, line 20: If the click event is prevented, won't the link be broken?

packages/marko-web-native-x/utils/gtm-events.js
```diff
@@ -0,0 +1,19 @@
+export const pageLoad = (window) => {
```
zarathustra323: I don't think it's necessary to pass the `window` object to these utilities since it's a global.
Or, if there's concern accessing the global, pass the `dataLayerNativeX` value instead

zarathustra323 (changes requested):

packages/marko-web-native-x/browser/track-outbound-links.vue
solocommand: The NativeX GTM container does some kind of nonsense here, the `outbound_click` event being pushed into the datalayer causes the navigation

solocommand:
[Embed Native-X stories within the site](https://trello.com/c/rHbOMWJt/48-embed-native-x-stories-within-the-site)

packages/marko-web-native-x/components/story/track-end-of-content.marko
```diff
@@ -0,0 +1,19 @@
+import defaultValue from "@parameter1/base-cms-marko-core/utils/default-value";
+import { get, getAsObject } from "@parameter1/base-cms-object-path";
+
+$ const { req } = out.global;
+$ const story = getAsObject(input, "story");
+$ const push = defaultValue(input.push, false);
+
+<if(push)>
+  <marko-web-gtm-push name="dataLayerNativeX" data={
```
zarathustra323: This should probably be it's own component should be explicitly called by the user before calling any of these tracking components

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #63, opened 2021-04-08
Create hook support and add `onAuthenticationSuccess`

zarathustra323:
Adds support for registering hook functions. Currently, only `onAuthenticationSuccess` has been added. Hooks can be registered with the `IdentityXConfiguration` object:

```js
// your-identity-x-config.js
const IdentityXConfiguration = require('@parameter1/base-cms-marko-web-identity-x/config');

const config = new IdentityXConfiguration({
  appId: 'your-app-id',
  onHookError: (e) => {
    // handle/log hook errors here...
  },
});
config.addHook({
  name: 'onAuthenticationSuccess',
  shouldAwait: true, // whether async functions or promises should be awaited. false by default
  fn: async ({ user, authToken, service }) => {
    // do stuff with the data. the `service` parameter is the `IdentityX` service instance.
    console.log({ user, authToken, service });
  },
});
module.exports = config;
```

Multiple functions can be registered for the same hook queue:
```js
config.addHook({
  name: 'onAuthenticationSuccess',
  fn: async ({ user, authToken, service }) => {
    console.log('hook 1');
  },
}).addHook({
  name: 'onAuthenticationSuccess',
  fn: async ({ user, authToken, service }) => {
    console.log('hook 2');
  },
});
```

### DIFF
--- a/packages/marko-web-identity-x/config.js
+++ b/packages/marko-web-identity-x/config.js
@@ -1,4 +1,7 @@
 const { get, getAsArray, getAsObject } = require('@parameter1/base-cms-object-path');
+const { isFunction: isFn } = require('@parameter1/base-cms-utils');
+
+const validHooks = ['onAuthenticationSuccess'];
 
 class IdentityXConfiguration {
   /**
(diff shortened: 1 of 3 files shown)


zarathustra323:

brandonbk (approved):

PR #65, opened 2021-04-12
Create Omeda+IdentityX package w/rapid identify

zarathustra323:
Allows for “rapid identification” of an IdentityX user with Omeda. Once identified, the Omeda IDs are added to the IdentityX user as external entity IDs. Both the “standard” and “encrypted” IDs are applied.


zarathustra323:

packages/marko-web-omeda-identity-x/rapid-identify.js
```diff
@@ -0,0 +1,91 @@
...
+    email: appUser.email,
+    ...(givenName && { firstName: givenName }),
+    ...(familyName && { lastName: familyName }),
+    ...(organization && { companyName: organization }),
+    ...(organizationTitle && { title: organizationTitle }),
+    ...(alpha3 && { countryCode: alpha3 }),
+    ...(regionCode && { regionCode }),
+    ...(postalCode && { postalCode }),
+  };
+  const { data } = await omedaGraphQL.mutate({
```
solocommand, line 72: Rather than doing 3 queries to two graph servers, why not do single query to the omeda graphql and have it kick off the external user id additions?

zarathustra323: Because the Omeda GraphQL server has nothing to do with converting IdentityX users. The Omeda API is "pure" in that it only handles direct Omeda operations. Something else needs to be responsible for converting the IdentityX user. That's what this is.

solocommand (approved):

PR #66, opened 2021-04-14
Create Omeda+IdentityX Rapid Identification Router

zarathustra323:
Can be used by a website to mount a JSON endpoint for rapidly identifying the active IdentityX user (if present).

If the Omeda ID for the configured brand is already set on the IdentityX user, the value will simply be passed through. If not set, the "rapid identification dance" is initiated.


brandonbk (approved):

zarathustra323:

PR #73, opened 2021-04-19
Create p1 Algolia site search package

B77Mills:
This is a port from the following [PR](https://github.com/parameter1/ac-business-media-websites/pull/277) to move the package to base-cms.  

This package will require env variable for the following 
 - ALGOLIA_API_KEY
 - ALGOLIA_APP_ID
 - ALGOLIA_DEFAULT_INDEX
 
 onStart the following middleware will also have to be accounted for 
``` 
const algoliaSiteSearch = require('@parameter1/marko-web-algolia-site-search/middleware/algolia');
const algoliaSiteSearchPaginated = require('@parameter1/marko-web-algolia-site-search/middleware/paginated');
 
 onStart: (app) => {
    app.use(algoliaSiteSearch());
    app.use(algoliaSiteSearchPaginated());
  },
  ```

### CODE REVIEW

File: packages/marko-web-algolia-site-search/components/index.marko [hunk 1]
```diff
@@ -0,0 +1,153 @@
+import { getAsObject, getAsArray, get } from "@parameter1/base-cms-object-path";
```
zarathustra323: This is probably too site specific to be included in a generic package... really this package should only provide the middleware and the query wrapper components (for doing the algolia query + the base cms query)

packages/marko-web-algolia-site-search/components/nodes/content-list.marko
```diff
@@ -0,0 +1,141 @@
+import { getAsObject, get } from "@parameter1/base-cms-object-path";
```
zarathustra323: Same here... this should be handled by the site

packages/marko-web-algolia-site-search/components/pagination-controls.marko
```diff
@@ -0,0 +1,53 @@
+import { cleanPath } from "@parameter1/base-cms-utils";
```
zarathustra323, line 1: This could be included, but it should be a "functional" component only -- meaning, the site should determine how the controls are rendered (via a `input.renderBody` call) and the package component merely provides the wrapping logic

packages/marko-web-algolia-site-search/graphql/fragments/content-list.js
```diff
@@ -0,0 +1,67 @@
+const gql = require('graphql-tag');
```
zarathustra323: Same for fragments -- these should be provided by the websites, as each site will likely have different field requirements depending on their layouts

packages/marko-web-algolia-site-search/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@ac-business-media/marko-web-algolia-site-search",
+  "description": "Algoli site search components for BaseCMS websites.",
```
zarathustra323: Missing the `a` in Alogila :)

packages/marko-web-algolia-site-search/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "version": "0.0.0",
```
zarathustra323: This version number should be set to the current value found in the root `lerna.json` file

zarathustra323 (changes requested):

PR #77, opened 2021-05-19
Allow finding lowercased from addresses

solocommand:
In advance of changing all `from` db values to lowercase, allow looking up both the presented value and lowercase version.


solocommand:
[Force redirects to be case insensitive](https://trello.com/c/Q3vUloTg/23-force-redirects-to-be-case-insensitive)

brandonbk (approved):

PR #78, opened 2021-05-19
Https support

solocommand:
Adds support and basic documentation for using a self-signed certificate with the `marko-web` package.

image

### DIFF
--- a/packages/marko-web/README.md
+++ b/packages/marko-web/README.md
@@ -1,0 +1,60 @@
+base-cms-marko-web
+===
+This package provides the core MarkoJS and Express components for building a website using the BaseCMS ecosystem.
+
+## Getting started
+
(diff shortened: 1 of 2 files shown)


zarathustra323 (commented):
What was the impetus for adding HTTPS here?

solocommand:
> What was the impetus for adding HTTPS here?

Needed to troubleshoot how some third-party code would work when running in an insecure context (https on the page, loaded via http)

solocommand:
[Update legacy insecure resources to use secure versions (Mixed Content warnings)](https://trello.com/c/UTT8UqTj/1-update-legacy-insecure-resources-to-use-secure-versions-mixed-content-warnings)

zarathustra323:
@solocommand Let's discuss in more detail on a call

PR #79, opened 2021-05-20
Update to elements that appear on the leaders page

Shinsina:
Screen Shot 2021-05-20 at 10 21 53 AM

This is done in order to properly translate all elements on the leaders/companies page(s) in Mundo.

### DIFF
--- a/packages/marko-web-inquiry/browser/default-form.vue
+++ b/packages/marko-web-inquiry/browser/default-form.vue
@@ -5,7 +5,10 @@
     <div class="row">
       <div class="col-md-6">
         <div class="form-group">
-          <form-label id="inquiry-form.first-name" :required="true">
+          <form-label v-if="labels.firstName" id="inquiry-form.first-name" :required="true">
+            {{ labels.firstName }}
+          </form-label>
+          <form-label v-else id="inquiry-form.first-name" :required="true">
             First Name
           </form-label>
           <input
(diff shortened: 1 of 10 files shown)


Shinsina:
[How to solve for Mundo Leaders entries in Spanish and English](https://trello.com/c/oKzezqPm/50-how-to-solve-for-mundo-leaders-entries-in-spanish-and-english)

zarathustra323 (changes requested):
Preferably this would use a global languages object that contains all the values per language code (`es`, for example). That object would contain the proper translations and a helper function could be used throughout to translate the elements, based on a global language setting. This would remove the need for all the if/else blocks and default values

solocommand:
> Preferably this would use a global languages object that contains all the values per language code (`es`, for example). That object would contain the proper translations and a helper function could be used throughout to translate the elements, based on a global language setting. This would remove the need for all the if/else blocks and default values

For an example of something like this, see the `i18n` function that was added to mundo: https://github.com/parameter1/pmmi-media-group-websites/pull/197/files/1d0340cdb9f02d9f97f762cd8b5f95171a24d782..99d67f920f7bb16118ed92b95d853d7d42e60004

Shinsina:
> > Preferably this would use a global languages object that contains all the values per language code (`es`, for example). That object would contain the proper translations and a helper function could be used throughout to translate the elements, based on a global language setting. This would remove the need for all the if/else blocks and default values
> 
> For an example of something like this, see the `i18n` function that was added to mundo: https://github.com/parameter1/pmmi-media-group-websites/pull/197/files/1d0340cdb9f02d9f97f762cd8b5f95171a24d782..99d67f920f7bb16118ed92b95d853d7d42e60004

Mundo does already use i18n the issue here was with these templates/components being deeply nested in Base and us currently not having other customers requiring us to do this sort of thing, I had been told to "go somewhere in the middle" in terms of how much engineering went into the solution, this could definitely be improved but as it stands it "meets the ask".

PR #81, opened 2021-05-20
Implement of i18n on inquiry form

Shinsina:


Shinsina:
[How to solve for Mundo Leaders entries in Spanish and English](https://trello.com/c/oKzezqPm/50-how-to-solve-for-mundo-leaders-entries-in-spanish-and-english)

packages/marko-web-inquiry/translations.js
```diff
@@ -0,0 +1,10 @@
+const translations = {
+  en: {
+    firstNameLabel: 'First Name', surnameLabel: 'Last Name', emailLabel: 'Email Address', phoneLabel: 'Phone Number', companyLabel: 'Company Name', jobTitleLabel: 'Job Title', countryLabel: 'Country', countryPlaceholder: 'Select country...', zipLabel: 'ZIP/Postal Code', commentsLabel: 'Comment', submitLabel: 'Submit',
```
solocommand: Add line breaks after each `,`

packages/marko-web-inquiry/translations.js
```diff
@@ -0,0 +1,10 @@
...
+  },
+  es: {
+    firstNameLabel: 'Primer nombre', surnameLabel: 'Apellidos', emailLabel: 'Correo electrónico', phoneLabel: 'Número telefónico', companyLabel: 'Nombre de compañía', jobTitleLabel: 'Título de trabajo', countryLabel: 'País', countryPlaceholder: 'Selecciona un país...', zipLabel: 'Código postal', commentsLabel: 'Comentarios', submitLabel: 'Enviar',
+  },
+};
+
+export default translations;
```
solocommand: When exporting a single item, don't define it as a variable (unless you need to re-use it elsewhere in the file). This should read
```js
export default {
  en: {
  // ...
```

packages/marko-web-inquiry/i18n.js
```diff
@@ -0,0 +1,15 @@
...
+  if (keys.includes(lang)) {
+    value = lang[fieldLabelKey];
+  } else {
+    throw new Error('LANGUAGE STRING MUST BE PROVIDED');
+  }
+  return value;
+}
+
+
+export default i18n;
```
solocommand: Same here, export the object directly instead of defining it first.

packages/marko-web-inquiry/i18n.js
```diff
@@ -0,0 +1,15 @@
+import translations from './translations';
+
+let value = '';
```
solocommand: (Almost) never use let.

packages/marko-web-inquiry/i18n.js
```diff
@@ -0,0 +1,15 @@
...
+function i18n(lang, fieldLabelKey) {
+  const keys = Object.keys(translations);
+  if (keys.includes(lang)) {
+    value = lang[fieldLabelKey];
+  } else {
+    throw new Error('LANGUAGE STRING MUST BE PROVIDED');
```
solocommand: Errors should give context to the developer of what caused the condition. In this case, provide the language and key that should have been defined. Something like this:
```
  if (!translations[lang]) {
    throw new Error(`No translations available for requested language ${lang}!`);
  }
  if (!translations[lang][key]) {
    throw new Error(`No translation available in ${lang} for requested key ${key}!`);
  }
  return translations[lang][key];
```

packages/marko-web-inquiry/i18n.js
```diff
@@ -0,0 +1,15 @@
+import translations from './translations';
+
+let value = '';
+function i18n(lang, fieldLabelKey) {
```
solocommand: Use ES6 fat arrow syntax rather than named functions:
`(lang, fieldLabelKey) => { ... };`

packages/marko-web-inquiry/browser/fields/country.vue
```diff
@@ -785,5 +786,10 @@ export default {
       default: false,
     },
   },
+  computed: {
```
solocommand: Don't use computed for this -- if you're going to wrap the translator function, use `methods`:
```
  methods: {
    translate(key) {
      return i18n(this.lang, key);
    },
  },
```

solocommand (changes requested):

packages/marko-web-inquiry/translations.js
zarathustra323: @solocommand This should have been caught by the linter and failed the build... any thoughts why it didn't?

packages/marko-web-inquiry/i18n.js
```diff
@@ -0,0 +1,12 @@
+import translations from './translations';
+
+export default (lang, fieldLabelKey) => {
+  const keys = Object.keys(translations);
+  if (!keys.includes(lang)) {
```
solocommand: One of my comments got lost! You don't need to use Object.keys here -- it's an object, so just check if the key is truthy using array syntax (`translations[lang]`) instead of converting to an array to check if the key is present.

solocommand (changes requested):

packages/marko-web-inquiry/browser/fields/country.vue
```diff
@@ -785,5 +786,10 @@ export default {
       default: false,
     },
   },
+  methods: {
+    resolveFieldLabel(label) {
+      return i18n(this.lang, label);
```
zarathustra323: `this.lang` is going to be undefined here.

zarathustra323 (changes requested):

packages/marko-web-inquiry/browser/default-form.vue
```diff
@@ -182,6 +187,9 @@ export default {
     comments: '',
   }),
   methods: {
+    resolveFieldLabel(label) {
```
zarathustra323: Semantically the method would read better as `translate(key)`

zarathustra323 (changes requested):

solocommand:
One last thing, can you throw screenshots of the form from PMMI, one from mundo and one from AW?

packages/marko-web-inquiry/browser/fields/country.vue
```diff
@@ -784,6 +785,15 @@ export default {
       type: Boolean,
       default: false,
     },
+    lang: {
+      type: String,
```
zarathustra323, line 789: This won't actually respect a non-default value unless you pass the `lang` prop from the parent `default-form` component. :)

zarathustra323 (changes requested):

Shinsina:
Screen Shot 2021-05-20 at 4 11 31 PM
Screen Shot 2021-05-20 at 4 18 51 PM

packages/marko-web-inquiry/components/form.marko
```diff
@@ -37,7 +38,8 @@ $ const { RECAPTCHA_SITE_KEY } = require('../env');
       formClass: `${blockName}__form`,
       contentId: content.id,
       contentType: content.type,
-      sitekey: RECAPTCHA_SITE_KEY
+      sitekey: RECAPTCHA_SITE_KEY,
+      lang: lang
```
solocommand: Missing trailing comma

packages/marko-web-inquiry/components/form.marko
```diff
@@ -10,6 +10,7 @@ $ const withDescription = defaultValue(input.withDescription, true);
 
 $ const blockName = "inquiry-form";
 $ const { RECAPTCHA_SITE_KEY } = require('../env');
+$ const lang = input.lang
```
solocommand: Missing semicolon

packages/marko-web-inquiry/browser/fields/country.vue
```diff
@@ -13,7 +13,7 @@
       @change="$emit('input', $event.target.value)"
     >
       <option disabled="disabled" value="">
-        {{ translate(countryPlaceholder) }}
+        {{ translate('countryPlaceholder') }}
```
solocommand: Use double quotes instead of single quotes in templates (Vue and Marko)

solocommand (changes requested):

PR #82, opened 2021-05-21
Add missing menu toggle button props

brandonbk:


packages/marko-web-theme-default/components/marko.json
```diff
@@ -35,6 +35,14 @@
       "type": "string",
       "default-value": "three-bars"
     },
+    "@initially-expanded":  {
+      "type": "boolean",
+      "default-value": false
+    },
+    "@buttonLabel": {
```
zarathustra323: This should be dasherized

zarathustra323 (changes requested):

packages/marko-web-theme-default/components/marko.json
brandonbk: Fixed.

zarathustra323 (approved):

PR #87, opened 2021-05-27
Add docs to `AllPublishedContentQueryInput`

solocommand:


services/graphql-server/src/graphql/definitions/platform/content/index.js
```diff
@@ -284,23 +287,71 @@ input ContentSitemapNewsUrlsQueryInput {
 }
 
 input AllPublishedContentQueryInput {
+  """
```
zarathustra323: Change all of these to inline strings please... i.e.`"Limits results to content with a primary site matching the current site context"`

zarathustra323 (changes requested):

PR #93, opened 2021-06-01
Use "now" for `unpublished` date criteria

B77Mills:
It was using since if set, but it should always use now. Content can be queried with publishedDate between two dates, but unpublished should always be before now.

### DIFF
--- a/packages/utils/src/get-published-content-criteria.js
+++ b/packages/utils/src/get-published-content-criteria.js
@@ -10,7 +10,8 @@ module.exports = ({
   after,
   primarySite,
 } = {}) => {
-  const date = since || new Date();
+  const now = new Date();
+  const date = since || now;
   const types = isArray(contentTypes) && contentTypes.length
     ? contentTypes : getDefaultContentTypes();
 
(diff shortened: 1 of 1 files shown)


packages/utils/src/get-published-content-criteria.js
```diff
@@ -30,7 +31,7 @@ module.exports = ({
     $and: [
       {
         $or: [
-          { unpublished: { $gt: date } },
+          { unpublished: { $gt: now } },
```
solocommand, line 34: I don't think this is correct behavior. With this change, you can no longer use the `since` parameter since the `date` var is no longer used for the unpublished date bound.

You'll need to find a balance here that allows the user to specify the unpublished date, rather than defaulting to `now`. Perhaps chaining an `$and` if `since` exists?

solocommand (changes requested):

packages/utils/src/get-published-content-criteria.js
zarathustra323: I disagree (I think)... the intent is to prevent expired content from showing... ever. It is an `allPublishedContent` operation, after all. :)

zarathustra323: How would you envision using an `$and` when `since` is present? With `now`? That would result in the exact same query result as written in this PR

zarathustra323 (approved):

PR #100, opened 2021-06-09
Add support for Google News Publication name

B77Mills:
- Add the ability to pass in Publication name instead of just getting it off of the site product.  
 - Add xmlEntities encoding on Title, ImageTitle & ImageCaption fields in the googleNewsSitemap xml


solocommand (dismissed):

services/sitemaps/src/routes/google-news.js
```diff
@@ -48,13 +49,14 @@ const createUrl = (website, {
   published,
   publication,
   images,
-}) => {
+}, publicationName) => {
   // News requires a publication, a published date and a title.
-  if (!publication || !published || !title) return null;
+  if ((!publication && !publicationName) || !published || !title) return null;
+  const name = (publicationName) || publication.name;
```
zarathustra323: Whats the reason for needing to pass this in independently?

zarathustra323 (changes requested):

services/sitemaps/src/routes/google-news.js
solocommand: The google news publisher name needs to be able to be passed separate from the website name (CCJ vs Commercial Carrier Journal, for example). But that said, why not pass it as part of the object instead of adding a third parameter to this function call?

B77Mills: I can update

`.setUrls(contentSitemapNewsUrls.map(url => createUrl(website, url, publicationName)));`
 
to directly update the url.publication.name field when it maps the results or pass this as a third param.  Just not sure which version is best practice.  @zarathustra323 / @solocommand  preference?

solocommand: @B77Mills so, you are retrieving the `sitemapUrls` from the graph and you want to use the `publicationName` that was passed in in place of the URLs returned publication (website) name field?

Ultimately I don't think it makes sense to modify the incoming data and would probably leave it as-is. Maybe to make this slightly more clear, don't spread the item in the `createUrl` definition:
```diff
- const createUrl = (website, {
-   loc,
-   title,
-   published,
-   publication,
-   images,
- }) => {},
- publicationName) => {
+ const createUrl = (website, url, publicationName) => {
+ const {
+   loc,
+   title,
+   published,
+   publication,
+   images,
+ } = url;
+ const name = publicationName || publication.name;
```

B77Mills: Makes sense. That should be update.

services/sitemaps/src/routes/google-news.js
```diff
@@ -43,16 +43,17 @@ const createPublication = (name, language) => {
...
+  const {
+    loc,
+    title,
+    published,
+    publication,
+    images,
+  } = url;
+  const name = publicationName || publication.name;
   // News requires a publication, a published date and a title.
   if ((!publication && !publicationName) || !published || !title) return null;
```
solocommand: This can just be `if (!name || !published || !title) ...`

solocommand (changes requested):

zarathustra323:
Can we wait until tomorrow to discuss over the phone before merging?

solocommand (approved):

zarathustra323 (approved):

PR #104, opened 2021-06-10
Set default fallback value for lang if not set

Shinsina:
See https://github.com/parameter1/pmmi-media-group-websites/pull/228 . This creates a default fallback value if language is not set on the shared-inquiry-form component.

### DIFF
--- a/packages/marko-web-inquiry/components/form.marko
+++ b/packages/marko-web-inquiry/components/form.marko
@@ -11,7 +11,7 @@ $ const withDescription = defaultValue(input.withDescription, true);
 
 $ const blockName = "inquiry-form";
 $ const { RECAPTCHA_SITE_KEY } = require('../env');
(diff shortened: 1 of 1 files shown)


packages/marko-web-inquiry/components/form.marko
```diff
@@ -11,7 +11,7 @@ $ const withDescription = defaultValue(input.withDescription, true);
 
 $ const blockName = "inquiry-form";
 $ const { RECAPTCHA_SITE_KEY } = require('../env');
-$ const lang = input.lang;
+$ const lang = defaultValue(input.lang, "en")
```
solocommand: Missing ending semicolon.

solocommand (changes requested):

zarathustra323 (approved):

PR #108, opened 2021-06-17
Display image in native or cropped aspect ratio

B77Mills:
Change the fit from crop to fill and add fill=solid.  This makes it display the image based on height or width filling space first and fills the remainder with transparency.  Is logo causes it to fill with white instead of transparent.

native 16x9 || isLogo
Cropped Portrait || native portrait 

Screen Shot 2021-06-17 at 1 29 34 PM


zarathustra323 (approved):

B77Mills:
[Update lightbox to use fit=fill instead of crop](https://trello.com/c/2HA06ufP/79-update-lightbox-to-use-fitfill-instead-of-crop)

PR #109, opened 2021-06-18
Add ability to query for images that are approved for web and/or print

B77Mills:
Add the ability to query for content images that approved for web and print.  By default it will only return images that are approvedForWeb


B77Mills:
[Ensure only "Active Web" images are showing up in Lightbox (photoswipe)](https://trello.com/c/gjGcOm0a/80-ensure-only-active-web-images-are-showing-up-in-lightbox-photoswipe)

zarathustra323 (approved):

PR #115, opened 2021-07-15
Add `siteQueryOperator` to `WebsiteRedirectQueryInput`

zarathustra323:
When the `siteQueryOperator` input is set to `notEqual` the query when attempt to return a redirect that matches the `from` value from _any_ site other than the current site. This is _not_ the default behavior, and must be specifically set in the query input.


brandonbk (approved):

zarathustra323:
[Update content redirects to support cross-site lookups](https://trello.com/c/41gZGN8Q/134-update-content-redirects-to-support-cross-site-lookups)

PR #116, opened 2021-07-20
Add mutation to create a website schedule

solocommand:
The logic for this is largely the same as the `quickCreateWebsiteSchedules` with three notable exceptions:
- Supports setting additional fields (custom optionId, startDate, & endDate)
- Only allows a single schedule to be created at once
- Now requires GraphQL authentication

This mutation is being added to support Allured's external publishing tool. Due to the backwards incompatible changes to authentication, this mutation is being created alongside `quickCreateWebsiteSchedules`, rather than replacing it.

image

### DIFF
--- a/services/graphql-server/src/graphql/definitions/website/schedule.js
+++ b/services/graphql-server/src/graphql/definitions/website/schedule.js
@@ -18,6 +18,7 @@ extend type Query {
 }
 
 extend type Mutation {
+  createWebsiteSchedule(input: CreateWebsiteScheduleMutationInput!): WebsiteSchedule! @requiresAuth
   quickCreateWebsiteSchedules(input: QuickCreateWebsiteSchedulesMutationInput!): [WebsiteSchedule!]!
   updateWebsiteSchedule(input: UpdateWebsiteScheduleMutationInput!): WebsiteSchedule!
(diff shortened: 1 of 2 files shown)


solocommand:
[Create GraphQL mutation: CreateWebsiteSchedule](https://trello.com/c/WKznJhJM/102-create-graphql-mutation-createwebsiteschedule)

services/graphql-server/src/graphql/definitions/website/schedule.js
```diff
@@ -98,6 +99,11 @@ input WebsiteScheduleOptionInput {
   status: ModelStatus = active
 }
 
+input CreateWebsiteSchedulesMutationInput {
+  contentId: Int!
```
zarathustra323, line 103: Change to `definitions: [CreateWebsiteScheduleDefinitionInput!]!`

and then do something like
```graphql
input CreateWebsiteScheduleDefinitionInput {
  contentId: Int!
  sectionId: Int!
  optionId: Int
  startDate: Date
  endDate: Date
}
```

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #130, opened 2021-07-23
Show video icon if item has videos

Shinsina:
Screen Shot 2021-07-23 at 11 05 49 AM

Gormann-Rupp Pumps only had a playlistId on it but was previously not displaying the video icon.

### DIFF
--- a/packages/leaders-program/src/components/list/nav/contents.vue
+++ b/packages/leaders-program/src/components/list/nav/contents.vue
@@ -29,13 +29,17 @@ export default {
       type: Object,
       default: null,
     },
+    hasVideos: {
+      type: Boolean,
+      default: false,
+    },
   },
 
   computed: {
     showVideoIcon() {
       const { youtube } = this;
       if (!youtube) return false;
-      return Boolean(youtube.username || youtube.channelId);
+      return Boolean(youtube.username || youtube.channelId || this.hasVideos);
     },
     iconModifiers() {
       const mods = [];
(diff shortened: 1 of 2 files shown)


Shinsina:
[MINExpo - featured Exhibitors video icon](https://trello.com/c/wSyfCEnk/126-minexpo-featured-exhibitors-video-icon)

packages/leaders-program/src/components/list/nav/contents.vue
```diff
@@ -29,13 +29,17 @@ export default {
...
+      default: false,
+    },
   },
 
   computed: {
     showVideoIcon() {
       const { youtube } = this;
       if (!youtube) return false;
-      return Boolean(youtube.username || youtube.channelId);
+      return Boolean(youtube.username || youtube.channelId || this.hasVideos);
```
zarathustra323: @B77Mills and @Shinsina are the `username` and `channelId` checks even needed now? The `item.videos` property is being resolved using the content's `youtubeVideos` field, which would likely be empty if no `username` or `channelId` are present?

zarathustra323 (changes requested):

packages/leaders-program/src/components/list/nav/contents.vue
B77Mills: @zarathustra323   That is correct.  Then we can delete the computed showVideoIcon() and just update the the v-if on line 6 to `v-if="showVideos"`.  That also means you should replace the youtube prop with hasVideos instead of adding it.
```

Shinsina: @zarathustra323 Appears that they wouldn't. I went to check and make sure this didn't alter previously working entries on PackWorld and Automation World as well. I wasn't sure if there was going to be one where the channelId or username would be set and there wouldn't be any videos.

zarathustra323 (approved):

PR #133, opened 2021-07-30
Enable ability to disable card flyouts on leaders.

Shinsina:
Currently the default behavior of the leaders block to is have flyout blocks when a link is hovered over:

Screen Shot 2021-07-30 at 11 38 55 AM

This change allows for these flyout blocks to be disabled when allowFlyOut is set to false:

Screen Shot 2021-07-30 at 11 39 05 AM

This corresponds to https://github.com/parameter1/ascend-media-websites/pull/188 and the ticket attached to this PR.

### DIFF
--- a/packages/leaders-program/src/components/containers/section-wrapper.vue
+++ b/packages/leaders-program/src/components/containers/section-wrapper.vue
@@ -27,6 +27,7 @@
             :promotion-limit="promotionLimit"
             :video-limit="videoLimit"
             :featured-product-label="featuredProductLabel"
+            :allow-fly-out="allowFlyOut"
             @action="emitAction"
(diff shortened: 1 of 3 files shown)


Shinsina:
[MINExpo: Update "Featured Companies" block width on landing page](https://trello.com/c/TTLKiwOl/104-minexpo-update-featured-companies-block-width-on-landing-page)

zarathustra323:
@Shinsina This can already be controlled via the `open` config value. Setting open to `null` will disable the fly-out entirely. See https://github.com/parameter1/pmmi-media-group-websites/blob/master/packages/common/components/leaders/all.marko for an implementation of this.

PR #145, opened 2021-08-13
Allow null values for Youtube components

Shinsina:
It appears when a 400 level error response is returned by the Youtube API the GraphQL fragment isn't allowing the error to fall off and is displaying it's own error in the response as found here: https://parameter1.atlassian.net/browse/HELP-129 .

The easiest solution is to allow for null values to be set at the Graph level. I took into account how these were being rendered and it would appear those are all requiring the data to be set on them anyway so it won't cause an issue in which invalid data is being set as the data won't be there to be passed anyway. I made a test company to get an idea of what returns when bad data is passed (including a private playlist which was my primary testing point).

Potentially we should check if a particular field is present on the video items to validate if the object is properly defined or not, because even though Vue will continue to work passed "erroring" required properties we should avoid that whenever possible.

Screen Shot 2021-08-13 at 12 57 37 PM

### DIFF
--- a/services/graphql-server/src/graphql/definitions/google-data-api/youtube.js
+++ b/services/graphql-server/src/graphql/definitions/google-data-api/youtube.js
@@ -1,11 +1,13 @@
 const gql = require('graphql-tag');
 
 module.exports = gql`
+
 extend type Query {
   validateYoutubePlaylistId(input: ValidateYoutubePlaylistIdQueryInput!): Boolean!
   validateYoutubeChannelId(input: ValidateYoutubeChannelIdQueryInput!): Boolean!
   validateYoutubeUsername(input: ValidateYoutubeUsernameQueryInput!): Boolean!
 }
+
 enum YoutubeThumbnailSizes {
   default
   medium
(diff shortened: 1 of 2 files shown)


Shinsina:
[Improve error handling for leaders front end components](https://trello.com/c/qaT3N3np/94-improve-error-handling-for-leaders-front-end-components)

zarathustra323 (changes requested):
The error should be handled at the resolver level and these type definitions should not change.

zarathustra323:
Is the error because there is a video that's trying to be returned that's no longer found? If so, handle that on the Content.youtubeVideos resolver, not on the YouTube type resolvers

Shinsina:
> Is the error because there is a video that's trying to be returned that's no longer found? If so, handle that on the Content.youtubeVideos resolver, not on the YouTube type resolvers

The error is on video objects that aren't returning from the API correctly as they can't be found due to being 404 or 403 errored on the API call. Hence why it is presenting an error of "Cannot return null for non-nullable field youtubeVideo.thumbnail"

Edit: Moreover this occurs when valid data for playlistId, channelId or username are set meaning they have to be accounted for after the API call as while they are "valid" given all performable measures, they don't return valid data upon the API call.

Edit 2: It would appear there are validation queries that will return false for the data causing the issue but nothing is using them. I'll try to work a change that incorporates those so we can filter out that bad data before it gets to the rendering pipeline and errors out

PR #146, opened 2021-08-13
Modify resolver to return empty array on invalid resolutions for Youtube API

Shinsina:
Replaces: https://github.com/parameter1/base-cms/pull/145

### DIFF
--- a/packages/leaders-program/src/graphql/queries/content-for-section.js
+++ b/packages/leaders-program/src/graphql/queries/content-for-section.js
@@ -6,6 +6,7 @@ query ContentForLeadersSection(
   $sectionId: Int!,
   $promotionLimit: Int = 4,
   $videoLimit: Int = 3,
+  $showPrivate: false,
   $includeContentTypes: [ContentType!] = [Company]
 ) {
   websiteScheduledContent(input: {
@@ -107,7 +108,7 @@ query ContentForLeadersSection(
             channelId
             url
           }
-          videos: youtubeVideos(input: { pagination: { limit: $videoLimit } }) {
+          videos: youtubeVideos(input: { pagination: { limit: $videoLimit }, showPrivate: $showPrivate }) {
             edges {
               node {
                 id
--- a/packages/leaders-program/src/graphql/queries/content-for-section.js
+++ b/packages/leaders-program/src/graphql/queries/content-for-section.js
@@ -6,7 +6,6 @@ query ContentForLeadersSection(
   $sectionId: Int!,
   $promotionLimit: Int = 4,
   $videoLimit: Int = 3,
-  $showPrivate: false,
   $includeContentTypes: [ContentType!] = [Company]
 ) {
   websiteScheduledContent(input: {
@@ -108,7 +107,7 @@ query ContentForLeadersSection(
             channelId
             url
           }
-          videos: youtubeVideos(input: { pagination: { limit: $videoLimit }, showPrivate: $showPrivate }) {
+          videos: youtubeVideos(input: { pagination: { limit: $videoLimit } }) {
             edges {
               node {
                 id
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -659,27 +659,15 @@ module.exports = {
         maxResults,
         ...(pageToken && { pageToken }),
       };
-      const response = await googleDataApiClient.request('youtube.playlistItems', payload);
-      const find400 = [];
-      if (getAsArray(response, 'errors')) { // If the response contains errors
-        const errors = getAsArray(response, 'errors');
-        errors.forEach((error) => {
-          if (error.extensions.exception.statusCode >= 400) { // If there is a 400 error
-            find400.push(error);
-          }
-        });
+      try {
+        const response = await googleDataApiClient.request('youtube.playlistItems', payload);
+        return {
+          pageInfo: getAsObject(response, 'pageInfo'),
+          items: getAsArray(response, 'items').filter(video => (input.showPrivate ? video : get(video, 'status.privacyStatus') === 'public')),
+        };
+      } catch (e) {
+        return { pageInfo: {}, items: [], errors: e };
       }
-      if (find400.length > 0) {
-        return { pageInfo: {}, items: [] }; // If a 400 error is found return "nothing"
-      }
-      const initialVideoList = getAsArray(response, 'items');
-      const filteredVideoList = [];
-      initialVideoList.forEach((video) => {
-        if (video.status.privacyStatus === 'public') { // Check to make sure each video is public status
-          filteredVideoList.push(video);
-        }
-      });
-      return { pageInfo: getAsObject(response, 'pageInfo'), items: filteredVideoList };
     },
   },
 
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -52,7 +52,7 @@ const retrieveYoutubePlaylistId = async ({ youtube }) => {
   if (get(response, 'items.0.contentDetails.relatedPlaylists.uploads')) {
     return get(response, 'items.0.contentDetails.relatedPlaylists.uploads');
   }
-  return '';
+  return null;
 };
 
 const { isArray } = Array;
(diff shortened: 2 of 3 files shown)


Shinsina:
[Improve error handling for leaders front end components](https://trello.com/c/qaT3N3np/94-improve-error-handling-for-leaders-front-end-components)

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -49,7 +49,10 @@ const retrieveYoutubePlaylistId = async ({ youtube }) => {
     payload.forUsername = forUsername;
   }
   const response = await googleDataApiClient.request('youtube.channelList', payload);
-  return get(response, 'items.0.contentDetails.relatedPlaylists.uploads');
+  if (get(response, 'items.0.contentDetails.relatedPlaylists.uploads')) {
+    return get(response, 'items.0.contentDetails.relatedPlaylists.uploads');
+  }
+  return '';
```
brandonbk: @Shinsina - the title mentions returning an empty array, but this is returning an empty string...is that correct in this case?

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -659,7 +662,11 @@ module.exports = {
         maxResults,
         ...(pageToken && { pageToken }),
       };
-      return googleDataApiClient.request('youtube.playlistItems', payload);
+      const response = await googleDataApiClient.request('youtube.playlistList', { part: 'id', id: playlistId });
+      if (getAsArray(response, 'items').length > 0) {
+        return googleDataApiClient.request('youtube.playlistItems', payload);
```
brandonbk: @Shinsina - is it necessary to use two separate requests for this?

In other words, do you have to first call `youtube.playlistList` to determine if there are any results?  Or, can you figure that out with the second call and either return the response or an empty array accordingly?

brandonbk (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
Shinsina: @brandonbk This particular resolver/function returns the id of a playlist as a string as far as I know. The title of the PR is accurate in that what is actually being returned via API call resolution to the front-end Vue component will fall onto an empty array should the information be invalid.

services/graphql-server/src/graphql/resolvers/platform/content.js
Shinsina: @brandonbk This should now be improved in subsequent commits. I hadn't actually looked at how the responses were structured both responses return an "items" array and return the same data I was simply trying to reuse the logic found in existing GraphQL queries in order to do this validation, it should go to note however that as far as I can tell nothing is using these validation queries as the Leaders blocks call for the companies in blocks (I hadn't necessarily determined how the blocks were structured or the size of them but it's clear it calls for multiple at once) so the validation queries never have the chance to intervene.

Screen Shot 2021-08-16 at 9 23 49 AM

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -659,7 +662,11 @@ module.exports = {
         maxResults,
         ...(pageToken && { pageToken }),
       };
-      return googleDataApiClient.request('youtube.playlistItems', payload);
+      const response = await googleDataApiClient.request('youtube.playlistItems', payload);
+      if (getAsArray(response, 'items').length > 0) {
+        return response;
+      }
+      return [];
```
```suggestion
      return getAsArray(response, 'items');
```

solocommand (changes requested):
`null` is a valid response for the playlistId field, do not change it to an empty string. See discussion in #145, you need to adjust handling within the resolver to handle error conditions (404/401/etc), and in the front-end vue components to allow rendering parts of the response, such as a subset of videos.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -659,7 +662,27 @@ module.exports = {
...
+        const errors = getAsArray(response, 'errors');
+        errors.forEach((error) => {
+          if (error.extensions.exception.statusCode >= 400) { // If there is a 400 error
+            find400.push(error);
+          }
+        });
+      }
+      if (find400.length > 0) {
+        return { pageInfo: {}, items: [] }; // If a 400 error is found return "nothing"
+      }
```
solocommand: These errors should be returned to the client to handle. Further, the pageInfo and items should likely be returned as-is as well

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -659,7 +662,27 @@ module.exports = {
...
+      const initialVideoList = getAsArray(response, 'items');
+      const filteredVideoList = [];
+      initialVideoList.forEach((video) => {
+        if (video.status.privacyStatus === 'public') { // Check to make sure each video is public status
+          filteredVideoList.push(video);
+        }
+      });
+      return { pageInfo: getAsObject(response, 'pageInfo'), items: filteredVideoList };
```
solocommand: No need to define additional arrays here, just filter the response. Ideally this should be a input option to include private videos (defaulting to `false`).

```suggestion
      return {
        pageInfo: getAsObject(response, 'pageInfo'),
        items: getAsArray(response, 'items').filter((video) => showPrivate || get(video, 'status.privacyStatus') === 'public'),
      };
```

solocommand (changes requested):

Shinsina:
Here is a quick production vs the change this make comparison: 

This change:

Screen Shot 2021-08-16 at 3 33 15 PM

Production:
Screen Shot 2021-08-16 at 3 34 45 PM

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -659,7 +659,15 @@ module.exports = {
...
       };
-      return googleDataApiClient.request('youtube.playlistItems', payload);
+      try {
+        const response = await googleDataApiClient.request('youtube.playlistItems', payload);
+        return {
+          pageInfo: getAsObject(response, 'pageInfo'),
+          items: getAsArray(response, 'items').filter(video => (input.showPrivate ? video : get(video, 'status.privacyStatus') === 'public')),
+        };
+      } catch (e) {
+        return { pageInfo: {}, items: [], errors: e };
```
zarathustra323, line 669: This should check for the specific `e.message` (or some other identifier) to determine if the the error is indeed the one you want to catch. This ensures that other, legitimate errors are still thrown (in other words, you don't want to return this logic for _all_ errors). For example:

```js
if (e.message === 'some message you want to catch') return {};
throw e;
```

An empty object is sufficient here since the `YoutubePlaylistConnection` [resolvers](https://github.com/parameter1/base-cms/blob/c6c111e2cc386e28dc7f0c561ddb6a95864e1fe9/services/graphql-server/src/graphql/resolvers/google-data-api/youtube.js#L8) already handle the defaults.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -659,7 +659,15 @@ module.exports = {
         maxResults,
         ...(pageToken && { pageToken }),
       };
-      return googleDataApiClient.request('youtube.playlistItems', payload);
+      try {
+        const response = await googleDataApiClient.request('youtube.playlistItems', payload);
+        return {
```
zarathustra323, line 664: Spread the entire response and then override the `items` property. This way, the entire response object is preserved as-is. Otherwise other properties of the response object will be lost.

```js
return {
  ...(response || {}),
  items: getAsArray(response, 'items').filter(video => (input.showPrivate ? video : get(video, 'status.privacyStatus') === 'public')),
};
```

zarathustra323 (changes requested):

packages/leaders-program/src/graphql/queries/content-for-section.js
```diff
@@ -107,7 +108,7 @@ query ContentForLeadersSection(
             channelId
             url
           }
-          videos: youtubeVideos(input: { pagination: { limit: $videoLimit } }) {
+          videos: youtubeVideos(input: { pagination: { limit: $videoLimit }, showPrivate: $showPrivate }) {
```
zarathustra323: You need to actually define the `showPrivate` input on the `ContentCompany.youtubeVideos` definition. If you were to pass true to this input the GQL request would error because `showPrivate` isn't actually defined.

Also, is there a case in the leaders package where you're actually passing this in as `true`? I don't see that anywhere, so there really isn't a need to add it to the query.

That said, you _must_ add the input definition to the graphql server (as mentioned above).

Everything in GraphQL must be defined (types, inputs, enums, etc etc)... it does not allow for arbitrary values

zarathustra323 (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
zarathustra323: Also, the reason to use `...(response || {})` is because the spread operator allows for `undefined` and any object... however, a `null` value (or any other falsey value) would error. This converts the falsey value to an empty object to prevent that.

PR #148, opened 2021-08-18
Do not attempt deploys if images are already built

solocommand:
Prevents an older tag from being deployed when the Github > Travis webhook catchup happens (aka the 2am rollback to last weeks release).

If the image has already been built, assume it no longer needs to be deployed and pass the build without attempting to upgrade any containers.

This will remove the ability to rollback to an older version via Travis, but normal rollbacks within Rancher will be unaffected.


zarathustra323 (commented):
Preferably there would be a way to check which image version is deployed and base the logic on that but this will work for now

zarathustra323 (approved):

PR #155, opened 2021-09-07
Update load-more to include role main, add labels.

Shinsina:
Screen Shot 2021-09-07 at 9 11 50 AM

Consulted:
https://www.w3.org/TR/wai-aria-practices/examples/landmarks/main.html (Either the ARIA or HTML tab both seem to indicate essentially the same thing) 
https://www.w3.org/TR/wai-aria-practices/examples/landmarks/HTML5.html (defaulting labeling practices by tag, hence why a unique label per section would need to be provided regardless of role)

This will be used in conjunction with a forthcoming PR for Ascend's websites: https://github.com/parameter1/ascend-media-websites/pull/227


Shinsina:
[Load More Landmark:Main](https://trello.com/c/VkplomH7/178-load-more-landmarkmain)

zarathustra323 (approved):

PR #176, opened 2021-10-29
Embedded image attribute handling

solocommand:
- Move attribute defaults for embedded images to the GraphQL definitions for the relevant `body` field
- Change default `auto` parameter to include `compress`
- Change default `w` to be `1280`px instead of `1440`
- Include `src-set` with a pre-built 2x dpr tag with the embed tag

When specified, the `w`, `fit`, and `auto` parameters can be modified from the defaults by supplying input arguments to the `body` field (content or sidebar)

image

### DIFF
--- a/services/graphql-server/src/graphql/definitions/platform/content/index.js
+++ b/services/graphql-server/src/graphql/definitions/platform/content/index.js
@@ -175,7 +175,7 @@ enum RelatedContentQueryType {
 }
 
 type ContentStubSidebar {
-  body: String
+  body(input: ContentStubSidebarBodyInput = {}): String
   name: String
   label: String
   sequence: Int!
@@ -302,6 +302,15 @@ type QueryMostPopularContentConnection {
 
 type QueryMostPopularContentEdge {
   node: MostPopularContent!
+}
+
+input ContentStubSidebarBodyInput {
+  "Embedded image defaults to apply to inline images"
+  imageAttrs: EmbeddedImageAttrsInput = {
+    w: 1280,
+    fit: "max",
+    auto: "format,compress"
+  }
 }
 
 input ContentQueryInput {
--- a/services/graphql-server/src/graphql/definitions/platform/content/index.js
+++ b/services/graphql-server/src/graphql/definitions/platform/content/index.js
@@ -665,6 +665,20 @@ input ContentTeaserInput {
 
 input ContentBodyInput {
   mutation: ContentMutation = Website
+  imageAttrs: EmbeddedImageAttrsInput = {
+    w: 1280,
+    fit: "max",
+    auto: "format,compress"
+  }
+}
+
+input EmbeddedImageAttrsInput {
+  "The (max) width of the embedded image"
+  w: Int
+  "The Imgix 'fit' parameter"
+  fit: String
+  "The Imgix 'auto' parameter"
+  auto: String
 }
 
 input ContentTaxonomyInput {
--- a/services/graphql-server/src/graphql/definitions/platform/content/index.js
+++ b/services/graphql-server/src/graphql/definitions/platform/content/index.js
@@ -682,7 +682,7 @@ input ContentBodyInput {
 }
 
 input EmbeddedImageAttrsInput {
-  "The (max) width of the embedded image"
+  "The width of the embedded image"
   w: Int
   "The Imgix 'fit' parameter"
   fit: String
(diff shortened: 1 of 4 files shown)


services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -785,8 +785,10 @@ module.exports = {
   },
 
   ContentStubSidebar: {
-    body: async ({ body }, _, { site, basedb }) => {
-      const value = await prepareSidebarBody(body, { site, basedb });
+    body: async ({ body }, { input }, { site, basedb }) => {
+      const { imageAttrs } = input;
+      console.log('imageAttrs', imageAttrs);
```
zarathustra323: Remove log

services/graphql-server/src/graphql/utils/embedded-image-tags.js
```diff
@@ -20,16 +26,16 @@ module.exports = async (body, { imageHost, basedb }) => {
...
 
     tag.set('alt', createAltFor(image));
     tag.set('src', createSrcFor(imageHost, image, {
-      w: size,
-      fit: 'max',
-      auto: 'format',
+      ...defaults,
+      ...imageAttrs,
+    }));
+    tag.set('srcSet', createSrcFor(imageHost, image, {
```
zarathustra323: This is missing the 2x value required by the attribute

services/graphql-server/src/graphql/utils/embedded-image-tags.js
```diff
@@ -1,7 +1,13 @@
 const { extractEmbeddedTags } = require('@parameter1/base-cms-embedded-media');
 const { createAltFor, createSrcFor, createCaptionFor } = require('@parameter1/base-cms-image');
 
-module.exports = async (body, { imageHost, basedb }) => {
+const defaults = {
+  w: '1280',
+  fit: 'max',
+  auto: 'format,compress',
```
zarathustra323, line 7: Need to verify that the default compress of 45 doesn't make the image look like crap when not sending an enhanced format like avif or webp

services/graphql-server/src/graphql/definitions/platform/content/index.js
```diff
@@ -665,6 +674,20 @@ input ContentTeaserInput {
...
   mutation: ContentMutation = Website
+  imageAttrs: EmbeddedImageAttrsInput = {
+    w: 1280,
+    fit: "max",
+    auto: "format,compress"
+  }
+}
+
+input EmbeddedImageAttrsInput {
+  "The (max) width of the embedded image"
```
zarathustra323: This isn't actually the max width. It will force the image to the specified width regardless of the images actual size

zarathustra323 (changes requested):

services/graphql-server/src/graphql/utils/embedded-image-tags.js
brandonbk: @zarathustra323 - I've tested a handful; they're all coming back as JPEGs and have looked identical.

services/graphql-server/src/graphql/utils/embedded-image-tags.js
```diff
@@ -20,16 +26,16 @@ module.exports = async (body, { imageHost, basedb }) => {
...
-      w: size,
-      fit: 'max',
-      auto: 'format',
+      ...defaults,
+      ...imageAttrs,
+    }));
+    tag.set('srcSet', createSrcFor(imageHost, image, {
+      ...defaults,
+      ...imageAttrs,
+      dpr: 2,
```
brandonbk: Could consider changing this to `dpr: '2 2x'`

services/graphql-server/src/graphql/utils/embedded-image-tags.js
zarathustra323: @brandonbk Tested with auto formatting disabled?

services/graphql-server/src/graphql/utils/embedded-image-tags.js
zarathustra323: Can you also verify that the site will properly translate the src set from the tag to the HTML element?

services/graphql-server/src/graphql/utils/embedded-image-tags.js
zarathustra323: 2 is the correct attribute. 2x needs to be added to the end of the entire src set string value.

zarathustra323 (approved):

Issue #185, opened 2021-11-12
Incorrect results for `Content.websiteSchedules` when not including `sort`


solocommand:
Running the following query against leonis for `allured_all`:
```gql
query ContentExtractData($input: WebsiteScheduledContentQueryInput!) {
  websiteScheduledContent(input: $input) {
    totalCount
    edges {
      node {
        id
        name
        type
        ... on ContentCompany {
          websiteSchedules {
            option {
              name
            }
          }
        }
      }
    }
  }
}
```
with input
```json
{
  "input": {
    "sectionAlias": "directory-services",
    "optionName": "Pinned",
    "includeContentTypes": "Company",
    "includeLabels": [],
    "excludeLabels": []
  }
}
```
returns only a single result:
![image](https://user-images.githubusercontent.com/1778222/141504341-2c0fb629-dda7-4661-82b6-83a453e051a5.png)

Running the same query but including the `sort` parameter properly returns all expected schedules:
```json
{
  "input": {
    "sectionAlias": "directory-services",
    "optionName": "Pinned",
    "includeContentTypes": "Company",
    "includeLabels": [],
    "excludeLabels": [],
    "sort": {
      "field": "id"
    }
  }
}
```
![image](https://user-images.githubusercontent.com/1778222/141504625-3f536dec-07f1-4529-8de3-1f09d635e788.png)

Update this query to properly return all expected results without requiring the `sort` parameter to be passed.

PR #186, opened 2021-11-12
Remove default `{ field: null }` from `WebsiteScheduledContentQueryInput.sort`

Shinsina:
I'm not sure where in the resolving this causes the results to be limited to 1 but removing it resolves the issue documented in https://github.com/parameter1/base-cms/issues/185

Screen Shot 2021-11-12 at 11 51 29 AM


zarathustra323 (approved):

zarathustra323:
Resolves #185

PR #200, opened 2021-11-30
Add Redis caching to embed requests

zarathustra323:
Adds a Redis caching layer to embed requests, preventing constant re-queries of Embedly.

GET requests will set to, and read from, cache. If `no-cache` is present in the `cache-control` header, a fresh/non-cached version will be retrieved, set to cache, and returned (e.g. via Cmd+Shift+R).

POST requests will set to cache, but never read from it.

Note: the `REDIS_CACHE_DSN` must be set in production before deploying.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -381,6 +381,7 @@ services:
       <<: *env
       <<: *env-newrelic
       EMBEDLY_API_KEY: ${EMBEDLY_API_KEY-}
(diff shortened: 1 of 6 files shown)


zarathustra323:
[Add redis caching to oembed service](https://trello.com/c/aSNseXp3/33-add-redis-caching-to-oembed-service)

solocommand (approved):

PR #203, opened 2021-11-30
Enable use of HTML strings for byline field

Shinsina:
DEV:
Screen Shot 2021-11-30 at 2 41 13 PM
 
 
PROD: 
Screen Shot 2021-11-30 at 3 09 53 PM

DEV: 
Screen Shot 2021-11-30 at 3 08 57 PM

PROD:
Screen Shot 2021-11-30 at 3 09 01 PM

 I used FCP as a cross check to make sure this wasn't going to be breaking on other sites if there is potentially some sort of odd CSS targeting going on. It appears these are just using CSS ::before in order to populate the text and are using a text tag and rendering HTML as appropriate.
 
 Allured will need a dependency upgrade to enable this.

### DIFF
--- a/packages/marko-web/components/element/content/byline.marko
+++ b/packages/marko-web/components/element/content/byline.marko
@@ -10,5 +10,6 @@ import extractRender from "../extract-render";
   modifiers=input.modifiers
   attrs=input.attrs
   link=input.link
+  html=true
   ...extractRender(input)
 />


Shinsina:
[Need to support HTML tags in By Line](https://trello.com/c/spjO7AQ1/247-need-to-support-html-tags-in-by-line)

solocommand (approved):

PR #206, opened 2021-12-07
Allow for objects to be passed into `contentTypes` to set labels

Shinsina:
This allows for these to be only overridden as necessary. (As will be demonstrated in the PR that this one will be mentioned on).

Screen Shot 2021-12-07 at 12 07 51 PM


Shinsina:
[Change webinar to webcast on search filter](https://trello.com/c/HnddhVmD/195-change-webinar-to-webcast-on-search-filter)

packages/marko-web-search/config/index.js
```diff
@@ -44,12 +45,24 @@ class MarkoWebSearchConfig {
       assignedToWebsiteSectionIds: Joi.array().items(
         Joi.number().integer().min(1).required(),
       ).default([]),
+
+      contentTypeLabelMap: Joi.array().items(
```
zarathustra323: Instead, allow each item of `contentTypes` to either be a string (default) or an object with `type` and `label` properties. Then, when mapping these values, if the array item is a string, continue doing the default, otherwise set the id to using the object type and then set the label to what was passed

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #208, opened 2021-12-08
Add mutation to set custom attributes on companies

solocommand:
Allows for values to be nullable so they can be removed, if necessary.


solocommand:
[Modify Company Update form to support GSA# custom attribute](https://trello.com/c/EHOSCCbu/211-modify-company-update-form-to-support-gsa-custom-attribute)

zarathustra323 (approved):

PR #212, opened 2022-01-05
Add support for customBooleanFields to the IdentityX profile component

B77Mills:
**Will require: https://github.com/parameter1/identity-x/pull/6**

Screen Shot 2022-01-05 at 11 35 10 AM

### DIFF
--- a/packages/marko-web-identity-x/browser/profile.vue
+++ b/packages/marko-web-identity-x/browser/profile.vue
@@ -73,6 +73,23 @@
           />
         </div>
 
+        <div v-if="customBooleanFieldAnswers.length" class="row">
+          <div
+            v-for="fieldAnswer in customBooleanFieldAnswers"
+            :key="fieldAnswer.id"
+            class="col-12"
+          >
+            <custom-boolean
+              :id="fieldAnswer.id"
+
+              :message="fieldAnswer.field.label"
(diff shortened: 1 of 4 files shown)


B77Mills:
[How to add checkbox fields to Identity-X](https://trello.com/c/adTAKA4F/258-how-to-add-checkbox-fields-to-identity-x)

packages/marko-web-identity-x/browser/profile.vue
```diff
@@ -345,6 +371,10 @@ export default {
       }
     },
 
+    onCustomBooleanChange() {
+      this.user.customBooleanFieldAnswers = this.customBooleanFieldAnswers;
```
zarathustra323, line 379: I'm a bit confused by what's happening here... how is the change being applied if the array is just being reset to itself?

zarathustra323 (changes requested):

packages/marko-web-identity-x/browser/profile.vue
B77Mills: @zarathustra323  I was missing a chunk of this from my site code.  Sorry it should take a answerId now and set the value to the opposite of what it was and also set the user.customBooleanFieldAnsers to this.customBooleanAnswers

zarathustra323 (approved):

PR #217, opened 2022-01-13
Add export deployment tool

solocommand:
CLI for parameter1/ab-media-exports#2

### DIFF
--- a/packages/export-deployment-tool/package.json
+++ b/packages/export-deployment-tool/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "@parameter1/base-cms-export-deployment-tool",
+  "version": "0.0.0",
+  "description": "The deployment tool for use with @parameter1/base-cms-marko-exports",
+  "main": "src/index.js",
+  "repository": "https://github.com/parameter1/base-cms/tree/master/packages/export-deployment-tool",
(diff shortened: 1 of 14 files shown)


packages/export-deployment-tool/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@parameter1/base-cms-export-deployment-tool",
+  "version": "0.0.0",
```
zarathustra323: This should be set to the current version in `lerna.json`

packages/export-deployment-tool/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "@endeavorb2b/rancher2cli": "^1.0.5",
+    "envalid": "^6.0.2",
+    "node-fetch": "^2.6.6",
+    "yargs": "^15.4.1"
+  },
+  "bin": {
+    "deploy": "./src/index.js"
+  },
+  "devDependencies": {
+    "eslint": "^5.16.0",
```
zarathustra323: The dev dependencies can be removed. These are already installed at the root level

zarathustra323 (dismissed):

brandonbk (approved):

PR #228, opened 2022-01-29
Add marko-web-native-x-query component

brandonbk:
Allows one or more native-x ads to be returned for further processing/formatting.

Enables a dedicated native ad block like this one:
Screen Shot 2022-01-29 at 4 57 56 PM


brandonbk:
Closing; can just call native-x-fetch-elements directly.

PR #230, opened 2022-02-01
Add docs for Omeda+IdentityX package

solocommand:


packages/marko-web-omeda-identity-x/README.md
```diff
@@ -1 +1,68 @@
 # BaseCMS Marko Omeda+IdentityX Integrations
+Omeda+IdentityX components for BaseCMS/Marko websites.
+
+Based on IDP requirements, install ***only one*** of these packages:
```
zarathustra323: The `only one` statement is generally correct. However, you can still install omeda or identity-x as well, if the site needs to access anything directly from those packages

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #231, opened 2022-02-02
Add `ContentUserRegistration.isCurrentlyRequired` field

B77Mills:
Add logic to only apply isRequired and access levels when the start and end date for requires Registration is met.

### DIFF
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -530,19 +530,24 @@ module.exports = {
 
     userRegistration: (content) => {
       const requiresRegistration = get(content, 'mutations.Website.requiresRegistration');
-      if (!requiresRegistration) return { isRequired: false, siteIds: [], accessLevels: [] };
+      if (!requiresRegistration) {
+        return {
+          isRequired: false,
+          isCurrentlyRequired: false,
+          siteIds: [],
+          accessLevels: [],
+        };
+      }
 
       const requiresRegistrationOptions = getAsObject(content, 'mutations.Website.requiresRegistrationOptions');
       const requiresAccessLevels = get(content, 'mutations.Website.requiresAccessLevels');
       const { startDate, endDate } = requiresRegistrationOptions;
-      if (requiresRegistration && (startDate || endDate)) {
-        const now = new Date();
-        if ((startDate && startDate > now) || (endDate && endDate < now)) {
-          return { isRequired: false, siteIds: [], accessLevels: [] };
-        }
-      }
+      const now = new Date();
+      const isCurrentlyRequired = (!startDate || startDate <= now) && (!endDate || endDate >= now);
+
       const userRegistration = {
         isRequired: Boolean(requiresRegistration),
+        isCurrentlyRequired,
         startDate,
         endDate,
         siteIds: getAsArray(requiresRegistrationOptions, 'siteIds'),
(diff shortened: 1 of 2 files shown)


services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -535,6 +535,12 @@ module.exports = {
       const requiresRegistrationOptions = getAsObject(content, 'mutations.Website.requiresRegistrationOptions');
       const requiresAccessLevels = get(content, 'mutations.Website.requiresAccessLevels');
       const { startDate, endDate } = requiresRegistrationOptions;
+      if (requiresRegistration && (startDate || endDate)) {
```
zarathustra323: I don't believe you want to change this resolver... if you do, the management interface will no longer function

zarathustra323 (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
zarathustra323: If you're trying to determine if registration is required for gating, that should be handled elsewhere

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -310,6 +310,14 @@ module.exports = {
    *
    */
   ContentUserRegistration: {
+    isCurrentlyRequired: async (root) => {
+      if (!root.isRequired) return false;
+      if (!root.startDate && !root.endDate) return root.isRequired;
+      const now = new Date();
+      if (root.startDate && root.startDate <= now) return false;
+      if (root.endDate && root.endDate >= now) return false;
+      return root.isRequired;
```
solocommand: I think your logic is reversed here; you want to return true if there is a start date and it's in the past. You'd need to check the end date to be sure in that condition though. Something like this?
```suggestion
      if (root.startDate && root.startDate <= now) {
        if (root.endDate && root.endDate <= now) return false;
      }
      return true;
```

solocommand (dismissed):

services/graphql-server/src/graphql/resolvers/platform/content.js
B77Mills: @solocommand I did have the checks wrong, but I think they still are independent conditions.  I changed it to 

```
if (root.startDate && root.startDate > now) return false;
if (root.endDate && root.endDate < now) return false;
```

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -310,6 +310,14 @@ module.exports = {
    *
    */
   ContentUserRegistration: {
+    isCurrentlyRequired: async (root) => {
```
zarathustra323: Destructure `root` and update the ifs to use the vars directly...
```js
isCurrentlyRequired: async ({ isRequired, startDate, endDate }) => {
```

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -530,11 +538,18 @@ module.exports = {
 
     userRegistration: (content) => {
       const requiresRegistration = get(content, 'mutations.Website.requiresRegistration');
-      if (!requiresRegistration) return { isRequired: false, siteIds: [], accessLevels: [] };
```
zarathustra323: Theses changes can be discarded

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #233, opened 2022-02-03
Add fallback for non-primary site primary section.

Shinsina:
This will use the first section on the given site within the content items schedules and set it as the primary section.

Example article is one from Food Logistics on SDCE (https://www.foodlogistics.com/professional-development/training/news/22031062/gartner-inc-supply-chain-leaders-expect-permanent-hybrid-work-model-for-frontline-workers)

Screen Shot 2022-02-03 at 2 28 55 PM

Screen Shot 2022-02-03 at 2 28 47 PM
.

### DIFF
--- a/packages/canonical-path/src/platform/content.js
+++ b/packages/canonical-path/src/platform/content.js
@@ -22,7 +22,7 @@ const pathResolvers = {
     // @todo This should eventually account for secondary sites/sections.
     // For now load an alternate from schedules
     const sectionQuery = getAsArray(content, 'sectionQuery');
-    if (sectionQuery.length && !section) {
+    if (sectionQuery.length) {
       const currentSiteSections = sectionQuery.map((schedule) => {
         if (schedule.siteId === site.id()) return schedule;
         return null;
--- a/packages/canonical-path/src/platform/content.js
+++ b/packages/canonical-path/src/platform/content.js
@@ -1,5 +1,5 @@
 const { isFunction: isFn, cleanPath } = require('@parameter1/base-cms-utils');
-const { get } = require('@parameter1/base-cms-object-path');
+const { get, getAsArray } = require('@parameter1/base-cms-object-path');
 const { BaseDB } = require('@parameter1/base-cms-db');
 const { dasherize } = require('@parameter1/base-cms-inflector');
 
@@ -19,8 +19,20 @@ const pathResolvers = {
     };
     const section = await load('websiteSection', id, { alias: 1 }, query);
     // @todo This should eventually account for secondary sites/sections.
-    // For now load home when not found.
-    return section ? section.alias : 'home';
+    // For now load an alternate from schedules
+    const sectionQuery = getAsArray(content, 'sectionQuery');
+    if (sectionQuery.length && !section) {
+      const currentSiteSections = sectionQuery.map((schedule) => {
+        if (schedule.siteId === site.id()) return schedule;
+        return null;
+      }).filter(v => v);
+      if (currentSiteSections.length) {
+        const currentSiteSection = await load('websiteSection', currentSiteSections[0].sectionId, { alias: 1 }, query);
+        if (currentSiteSection) return currentSiteSection.alias;
+      }
+    }
+    // If the requested alternate could not be found.
+    return 'home';
   },
   primaryCategoryPath: async (content, { load }) => {
     const ref = BaseDB.get(content, 'mutations.Website.primaryCategory');
--- a/packages/canonical-path/src/platform/content.js
+++ b/packages/canonical-path/src/platform/content.js
@@ -1,8 +1,7 @@
-const { isFunction: isFn, cleanPath } = require('@parameter1/base-cms-utils');
+const { isFunction: isFn, cleanPath, getSectionFromSchedules } = require('@parameter1/base-cms-utils');
 const { get } = require('@parameter1/base-cms-object-path');
 const { BaseDB } = require('@parameter1/base-cms-db');
 const { dasherize } = require('@parameter1/base-cms-inflector');
-const getSectionFromSchedules = require('../../../utils/src/get-section-from-schedules');
 
 const pathResolvers = {
   id: content => content._id,
--- a/packages/canonical-path/src/platform/content.js
+++ b/packages/canonical-path/src/platform/content.js
@@ -21,8 +21,13 @@ const pathResolvers = {
     if (section) return section.alias;
     // @todo This should eventually account for secondary sites/sections.
     // For now load an alternate from schedules
-    const sectionFromSched = await getSectionFromSchedules({ content, siteId: site.id(), load });
-    if (sectionFromSched) return sectionFromSched.alias;
+    const sectionFromSchedule = await getSectionFromSchedules({
+      content,
+      siteId: site.id(),
+      projection: { alias: 1 },
+      load,
+    });
+    if (sectionFromSchedule) return sectionFromSchedule.alias;
     // If the requested alternate could not be found.
     return 'home';
   },
--- a/packages/canonical-path/src/platform/content.js
+++ b/packages/canonical-path/src/platform/content.js
@@ -18,7 +18,7 @@ const pathResolvers = {
       'site.$id': site.id(),
     };
     const section = await load('websiteSection', id, { alias: 1 }, query);
-    if (section) return section;
+    if (section) return section.alias;
     // @todo This should eventually account for secondary sites/sections.
     // For now load an alternate from schedules
     const sectionQuery = getAsArray(content, 'sectionQuery');
--- a/packages/canonical-path/src/platform/content.js
+++ b/packages/canonical-path/src/platform/content.js
@@ -1,7 +1,8 @@
 const { isFunction: isFn, cleanPath } = require('@parameter1/base-cms-utils');
-const { get, getAsArray } = require('@parameter1/base-cms-object-path');
+const { get } = require('@parameter1/base-cms-object-path');
 const { BaseDB } = require('@parameter1/base-cms-db');
 const { dasherize } = require('@parameter1/base-cms-inflector');
+const getSectionFromSchedules = require('../../../utils/src/get-section-from-schedules');
 
 const pathResolvers = {
   id: content => content._id,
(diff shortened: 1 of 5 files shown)


Shinsina:
[Content primary to food is showing to SDC as 'home'](https://trello.com/c/VweV2Fi2/167-content-primary-to-food-is-showing-to-sdc-as-home)

Shinsina:
[Why does "Home" show on some content bars.](https://trello.com/c/dRB9Y2tC/227-why-does-home-show-on-some-content-bars)

solocommand (changes requested):
@Shinsina This is great! One change -- use the `needs` parameter to the `@projection` directive on the definition and you won't need to do another `load`/query of the content.
image
In your resolver, just use `getAsArray` to retrieve the `sectionQuery` field from the `content` item.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -447,7 +447,7 @@ module.exports = {
     /**
      * Load primary section of content.
      * If primary section's site matches the current site, return the section.
-     * If not, check for alternative site + section (@todo).
+     * If not, check for alternative site + section.
```
zarathustra323: The todo comments should actually stay, as they are referring to explicitly allowing a user to set a secondary site/section, not a fallback to a schedule

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -482,7 +482,17 @@ module.exports = {
       if (section) return section;
 
       // Current section does not match site, load alternate.
-      // @todo This should eventually account for secondary sites/sections. For now, load home.
+      const sectionQuery = getAsArray(content, 'sectionQuery');
```
zarathustra323: This can be accomplished by returning _one_ section, not all of them. No need to query all possible sections when you're only returning one of them. :)

Instead, change your query to include the current siteId context, include where the schedule section Ids exist, and where the alias _is not_ home

zarathustra323 (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
zarathustra323: Something like:

```js
const currentSiteSectionQuery = {
  ...formatStatus(status),
  'site.$id': { $ne: siteId },
  _id: { $in: sectionQuery.map(schedule => schedule.sectionId) },
  alias: { $ne: 'home' },
};
```

Also, the if should also require that `siteId` be present. If no site context is provided then there's no way to know what to return

solocommand: > This can be accomplished by returning _one_ section, not all of them. No need to query all possible sections when you're only returning one of them. :)
> 
> Instead, change your query to exclude the current siteId context, include where the schedule section Ids exist, and where the alias _is not_ home

The current siteId context is correct at this point (it _should_ be included, not excluded), but agreed on `findOne` instead of `find` and the alias

zarathustra323 (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
zarathustra323: Yep, already fixed the siteId part

zarathustra323:
```js
const currentSiteSectionQuery = {
  ...formatStatus(status),
  'site.$id': siteId,
  _id: { $in: sectionQuery.map(schedule => schedule.sectionId) },
  alias: { $ne: 'home' },
};
```

solocommand: @Shinsina you can copy the website context exception from the `siteContext` field to ensure that a site id is present.

zarathustra323:
@solocommand and @Shinsina Ideally the `siteId` should be saved on each subdocument of the `Content.sectionQuery` array... if it were present, the resolver could filter the array ahead of time and load the section via a dataloader using the section ID (instead of doing a hard query). The benefit is that the dataloader would then be able to batch multiple requests... so a list of 50 content items would result in one query to the section collection, where right now this logic would result in 50 _individual_ new queries to get the fallback section (assuming all needed fallback)

Shinsina:
@solocommand and @zarathustra323 Additionally we'll need to consider doing the same thing here: https://github.com/parameter1/base-cms/blob/master/packages/canonical-path/src/platform/content.js#L21 so it would likely make sense to script something to add siteId to each subdocument of ```Content.sectionQuery``` (for already existing content) and modify the interface to additionally account for adding/handling that field when schedules are created. 

I think I managed to make sense of the previous comments in my most recent commit but if I missed something let me know.

zarathustra323:
@Shinsina yes, good point. The canonical path would also need to account for this... so... adjusting the data is probably necessary. Though such a change touches a lot of places. @solocommand thoughts?

Shinsina:
@zarathustra323 @solocommand I've corrected the resolution logic for primarySection to utilize siteId off the sectionQuery subdocuments and used a dataloader to grab the appropriate section, additionally I added (essentially) identical logic for the path resolution as well. 

Let me know if there is anything I missed or additional corrections that need to be made.

B77Mills (dismissed):

packages/canonical-path/src/platform/content.js
```diff
@@ -18,9 +18,22 @@ const pathResolvers = {
...
+    if (sectionQuery.length) {
+      const currentSiteSections = sectionQuery.map((schedule) => {
+        if (schedule.siteId === site.id()) return schedule;
+        return null;
+      }).filter(v => v);
+      if (currentSiteSections.length) {
+        const currentSiteSection = await load('websiteSection', currentSiteSections[0].sectionId, { alias: 1 }, query);
+        if (currentSiteSection) return currentSiteSection.alias;
+      }
+    }
```
solocommand: Use `Array.find` instead of map+filter when returning a single result.
```suggestion
    const item = sectionQuery.find(({ siteId }) => siteId === site.id());
    if (item) {
      const secondary = await load('websiteSection', item.sectionId, { alias: 1 }, query);
      if (secondary) return secondary.alias;
    }
```

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -482,7 +482,20 @@ module.exports = {
...
+        if (!siteId) throw new UserInputError('A site id must be set to generate the `Content.primarySection` field.');
+        const currentSiteSections = sectionQuery.map((schedule) => {
+          if (schedule.siteId === siteId) return schedule;
+          return null;
+        }).filter(v => v);
+        if (currentSiteSections.length) {
+          const currentSiteSection = await load('websiteSection', currentSiteSections[0].sectionId, projection, query);
+          if (currentSiteSection) return currentSiteSection;
+        }
+      }
```
solocommand: Same change here. Since you're doing the same thing in both files, export this logic as a utility function. It should take the `content` object and `load` function as parameters, and return the found section object (if present).

In the path resolver, you'd just return the alias from the secondary section, if it was found.

solocommand (changes requested):

packages/utils/src/get-section-from-schedules.js
```diff
@@ -0,0 +1,15 @@
+module.exports = async ({ content, siteId, load }) => {
+  const { sectionQuery } = content;
```
zarathustra323, line 7: Add `if (!sectionQuery.length) return null;` as the first line, then delete the `if` wrapper on L8 and the other `return null` on L12

packages/utils/src/get-section-from-schedules.js
```diff
@@ -0,0 +1,15 @@
...
+  const query = {
+    status: 1,
+    'site.$id': siteId,
+    alias: { $ne: 'home' },
+  };
+  if (sectionQuery.length) {
+    const sectionsFromScheds = sectionQuery.filter(schedule => schedule.siteId === siteId);
```
zarathustra323: These need to be stringify if they're MongoIds otherwise they'll never match.

`ObjectId()` does not equal `ObjectId()` because they are different object instances....

Change to
```js
`${schedule.siteId}` === `${siteId}`
```

packages/utils/src/get-section-from-schedules.js
```diff
@@ -0,0 +1,15 @@
...
+    const foundSections = await Promise.all(sectionsFromScheds.map(section => load('websiteSection', section.sectionId, {}, query)));
```
zarathustra323, line 10: Immediately bail if no sectionId is found, e.g. `if (!sectionId) return null`

Then do a _single_ load for the sectionId, not a Promise.all of many

zarathustra323: The `projection` argument should not be an empty object, otherwise all fields will return. Add the `projection` argument to this util function and pass it to the `load` function

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -482,7 +482,10 @@ module.exports = {
       if (section) return section;
 
       // Current section does not match site, load alternate.
-      // @todo This should eventually account for secondary sites/sections. For now, load home.
+      // @todo This should eventually account for secondary sites/sections.
+      // For now load an alternate from schedules.
+      const sectionFromSched = await getSectionFromSchedules({ content, siteId, load });
```
zarathustra323: For clarity, might as well name the variable the full name ... `sectionFromSchedule`

packages/canonical-path/src/platform/content.js
```diff
@@ -18,9 +18,13 @@ const pathResolvers = {
       'site.$id': site.id(),
     };
     const section = await load('websiteSection', id, { alias: 1 }, query);
+    if (section) return section.alias;
     // @todo This should eventually account for secondary sites/sections.
-    // For now load home when not found.
-    return section ? section.alias : 'home';
+    // For now load an alternate from schedules
+    const sectionFromSched = await getSectionFromSchedules({ content, siteId: site.id(), load });
```
zarathustra323: The projection here should be explicitly passed as `{ alias: 1 }`

services/graphql-server/src/graphql/resolvers/platform/content.js
zarathustra323: Pass `projection` through as an argument parameter

zarathustra323 (changes requested):

Shinsina:
@zarathustra323 Do we want it to just return the first section regardless of whether or not it is aliased as "home", say I have an article scheduled to home, a section called foo and a section called bar on the non-primary site, if the sectionQuery just so happens to have it in the order of Home > foo > bar should it return home or a "more specific" section?

packages/utils/src/get-section-from-schedules.js
Shinsina: I need all the fields as this is also being used to return the section itself here: https://github.com/parameter1/base-cms/pull/233/files/bf814341b5cb4a668917cb74833ab34425de5490#diff-466f539f3ec0e3fd7f2308c36f524ccf2ff92b8ef482f01cf0cf5d0eeb98e4faR488

zarathustra323: > Do we want it to just return the first section regardless of whether or not it is aliased as "home",

@Shinsina yes, you are correct... I forgot about the alias requirement. I'll adjust my comments where necessary

zarathustra323: @Shinsina the projection should be passed in by the callees, (see my comment on L487 of `resolvers/platform/content.js`)

zarathustra323: Also, I removed the comments asking to change the `filter` to `find`

zarathustra323 (approved):

solocommand (approved):

PR #237, opened 2022-02-08
Add `AssetImage.inCarousel` field and input

B77Mills:


B77Mills:
[Article slideshows have not transferred over](https://trello.com/c/Khw0uYuk/129-article-slideshows-have-not-transferred-over)

services/graphql-server/src/graphql/definitions/platform/asset.js
```diff
@@ -32,6 +32,7 @@ type AssetImage {
   caption: String @projection
   credit: String @projection
   cropDimensions: AssetImageCrop @projection
+  inCarousel: Boolean @projection
```
zarathustra323, line 35: Should this also be _is_ (instead of _in_) to match L106 below?

zarathustra323 (dismissed):

services/graphql-server/src/graphql/definitions/platform/asset.js
B77Mills: @zarathustra323 I just matched what is currently being saved on assets.  If we would prefer it to be `is`.  I can update it within management to the new field name and convert any data that has it already set.  

Thoughts?

zarathustra323: Then I'd leave it as `in` on the `Type` but change the `is` on the `Input` to `in` -- just so things are consistent

solocommand (approved):

PR #247, opened 2022-02-15
Addition of ```sectionStartDate``` field and field resolver.

Shinsina:
Screen Shot 2022-02-15 at 10 57 11 AM


Shinsina:
[Update GraphQL API to return current section scheduled date](https://trello.com/c/HgTodiyh/280-update-graphql-api-to-return-current-section-scheduled-date)

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -444,6 +444,26 @@ module.exports = {
       return isArray(sectionQuery) ? sectionQuery : [];
     },
 
+    /**
+     * Return the oldest start date for the current section
+     */
+    sectionStartDate: async (content, { input }, { basedb }) => {
+      const { sectionQuery } = content;
+      const { sectionId } = input;
+      const descendantIds = await getDescendantIds(sectionId, basedb);
```
solocommand: This should only happen if section bubbling is enabled. Add that to the input and only get the descendant ids in that case. Check the `websiteScheduledContent` query, match the input key and default value here as well (I believe on by default, but double check please!)

The rest of your logic should stay the same...something like `ids = useSectionBubbling ? await getDescendantIds() : [sectionId];`

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -444,6 +444,26 @@ module.exports = {
...
+      const descendantSectionSchedules = sectionQuery
+        .filter(section => descendantIds.includes(section.sectionId))
+        .map(section => section.start);
+      const sectionDates = [
+        ...new Set([
+          ...sectionSchedules.map(section => section.start),
+          ...descendantSectionSchedules,
+        ]),
+      ];
+      return sectionDates[0];
```
solocommand: We don't care about uniqueness, you should just be able to do something like this
```js
  return schedules.length ? schedules[0].start : null;
```

services/graphql-server/src/graphql/definitions/platform/content/interfaces/content.js
```diff
@@ -88,6 +88,8 @@ interface Content @requiresProject(fields: ["type"]) {
 
   # Allows accessing the content's taxonomy IDs without querying the related models.
   taxonomyIds: [Int!]! @projection(localField: "taxonomy")
+
```
solocommand, line 91: Add a GraphQL doc line above this field to indicate the expected use.

services/graphql-server/src/graphql/definitions/platform/content/index.js
```diff
@@ -778,6 +778,10 @@ input ContentHasWebsiteScheduleInput {
   sectionBubbling: Boolean = true
 }
 
+input SectionStartDateInput {
+  sectionId: Int
```
solocommand: Add GraphQL docs here to indicate what the input is used for and what each field (including the new flag) does

solocommand (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -444,6 +444,26 @@ module.exports = {
...
+     * Return the oldest start date for the current section
+     */
+    sectionStartDate: async (content, { input }, { basedb }) => {
+      const { sectionQuery } = content;
+      const { sectionId } = input;
+      const descendantIds = await getDescendantIds(sectionId, basedb);
+      const sectionSchedules = sectionQuery.filter(section => section.sectionId === sectionId);
+      const descendantSectionSchedules = sectionQuery
+        .filter(section => descendantIds.includes(section.sectionId))
+        .map(section => section.start);
```
solocommand: I wasn't explicit about it in my review, but combine these. There's no need to create two arrays just to join them again later. The website section ID is always included in its own descendant ids, but you can also just include it in your ids to check.
```js
const ids = [
  sectionId,
  ...(useSectionBubbling ? await getDescendantIds(...) : []),
];

zarathustra323:
@Shinsina we're going to hold on these changes for now, pending some other options

PR #248, opened 2022-02-15
Add city, street, extra fields, hidden support

solocommand:
Adds support for `city`, `street`, and `addressExtra` within IdentityX. Reorders form fields (slightly) to make for a more consistant layout when these fields are not displayed (the default)

| Description | Screenshot |
| - | - |
| Current (no selection) | image
| Current (NA) | image
| Current (non-NA) | image
--
| New (no selection) | image
| New with fields shown (NA) | image
| New with fields shown (non-NA) | image
--
| New with fields hidden (NA) | image
| New with fields hidden (non-NA) | image

### DIFF
--- a/packages/marko-web-identity-x/browser/profile.vue
+++ b/packages/marko-web-identity-x/browser/profile.vue
@@ -43,10 +43,11 @@
         </div>
 
         <address-block
-          v-if="countryCode"
+          v-if="showAddressBlock"
           :user="user"
           :is-field-required="isFieldRequired"
           :is-field-visible="isFieldVisible"
+          :display-region-field="displayRegionField"
         />
 
         <div v-if="customSelectFieldAnswers.length" class="row">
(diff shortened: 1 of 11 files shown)


solocommand:
@zarathustra323 field layout/logic updated as we discussed and new screenshots added.

packages/marko-web-identity-x/browser/profile.vue
```diff
@@ -317,17 +299,96 @@ export default {
...
+    },
+    organizationSettings() {
+      return {
+        required: this.isFieldRequired('organization'),
+        visible: this.isFieldVisible('organization'),
+      };
+    },
+    organizationTitleSettings() {
+      return {
+        required: this.isFieldRequired('organizationTitle'),
```
zarathustra323: Unfortunately can't use the function calls here either otherwise changes to the array won't trigger a change to the computed prop.

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #249, opened 2022-02-15
Place scheduled on sectionQuery, use startDate from schedule.

Shinsina:
Ref: https://github.com/parameter1/base-cms/pull/247


packages/section-query-generator/src/index.js
```diff
@@ -69,6 +69,7 @@ module.exports = async ({
             else: '$expires',
           },
         },
+        scheduled: '$startDate',
```
zarathustra323, line 72: What's this intended to be used for?

zarathustra323 (changes requested):

packages/section-query-generator/src/index.js
Shinsina: Currently there isn't a field on sectionQuery if the schedule has a startDate set prior to the content published date (why this is a "thing" beats me). This adds one.

Screen Shot 2022-02-15 at 2 40 29 PM
Screen Shot 2022-02-15 at 2 40 24 PM

zarathustra323: For what purpose? Section query intentionally uses published date to correct certain schedule start dates... in that example, that start date is actually wrong... you shouldn't be able to start content before it's actually published. Nothing in the interface prevents this, per se, hence why section query enforces this

Shinsina: https://trello.com/c/XpVjx5vG/277-incorrect-date-being-displayed-on-content-list-items-for-non-primary-sites

I don't agree with the idea of putting dates prior to the publish date anywhere.

zarathustra323:
@Shinsina we're going to hold on these changes for now, pending some other options

PR #253, opened 2022-02-24
Update to emit warning vs error when no site found

B77Mills:
emit a warning vs error to prevent the page from not loading.  This will just throw an error for each newsletter with no site assigned.  We could just remove this all together and only return newsletters with a site assigned.


jwade1327 (approved):

solocommand (approved):

solocommand:
[newsletters.oemoffhighway.com internal server error](https://trello.com/c/ZSKyhIL5/190-newslettersoemoffhighwaycom-internal-server-error)

PR #257, opened 2022-03-14
Expand Omeda promo code handling

solocommand:
This PR adds the following configuration options to the Omeda+IdentityX package. Each of these can be overridden by specifying a new value in the sites startServer handler.

@zarathustra323 I think this covers the conditions we talked about regarding sticky codes and site customization. Let me know if I missed something!

| Param | Default Value | Description
| ----------- | ----------- | ----------- |
| `omedaPromoCodeCookieName` | `omeda_promo_code` | The cookie/query parameter that should be checked for promo codes |
| `omedaPromoCodeDefault` | `undefined` | The promo code that should be set if there is none available from the request (cookie/query param) |

### omedaPromoCodeCookieName
The specified cookie will be set from the same named URL parameter, similar to the `oly_enc_id` handling. If this cookie is present on the request, the value will be used as the promo code sent with the rapid identification calls.

### omedaPromoCodeDefault
A global default promo code value to set, if the user has no specific promo code.

In practice, this means the following:
1. An anonymous visitor accessing a base-cms site with a URL parameter of `?omeda_promo_code=foo.bar.baz` will have the promo code `FOO.BAR.BAZ` set. If the visitor then attempts to log in via IdentityX (or returns later to do so), this promo code will be sent with the Omeda rapid identification call.
2. An authenticated visitor accessing the site with a new promo code will have the new promo code stored. Each request that rapidly identifies them in the future will include this promo code.
3. If a visitor returns to the site with a different promo code (`baz.bar.foo`), the _new_ promocode will be used and the original promo code will be discarded.

### DIFF
--- a/packages/marko-web-omeda-identity-x/index.js
+++ b/packages/marko-web-omeda-identity-x/index.js
@@ -16,7 +16,7 @@ module.exports = (app, {
   omedaGraphQLClientProp = '$omedaGraphQLClient',
   omedaRapidIdentifyProp = '$omedaRapidIdentify',
 
-  omedaPromoCodeCookieName = 'oly_promo_src',
+  omedaPromoCodeCookieName = 'omeda_promo_code',
   omedaPromoCodeStripQueryParam = true,
   omedaPromoCodeDefault,
 
@@ -72,7 +72,7 @@ module.exports = (app, {
   // strip `oly_enc_id` when identity-x user is logged-in
   app.use(stripOlyticsParam());
 
-  // set `oly_promo_src` when the URL parameter is present
+  // set `omeda_promo_code` when the URL parameter is present
   app.use(setPromoSourceCookie({
     omedaPromoCodeCookieName,
     omedaPromoCodeStripQueryParam,
--- a/packages/marko-web-omeda-identity-x/index.js
+++ b/packages/marko-web-omeda-identity-x/index.js
@@ -17,7 +17,6 @@ module.exports = (app, {
   omedaRapidIdentifyProp = '$omedaRapidIdentify',
 
   omedaPromoCodeCookieName = 'omeda_promo_code',
-  omedaPromoCodeStripQueryParam = true,
   omedaPromoCodeDefault,
 
   idxConfig,
@@ -75,6 +74,5 @@ module.exports = (app, {
   // set `omeda_promo_code` when the URL parameter is present
   app.use(setPromoSourceCookie({
     omedaPromoCodeCookieName,
-    omedaPromoCodeStripQueryParam,
   }));
 };
--- a/packages/marko-web-omeda-identity-x/index.js
+++ b/packages/marko-web-omeda-identity-x/index.js
@@ -1,6 +1,7 @@
 const omeda = require('@parameter1/base-cms-marko-web-omeda');
 const identityX = require('@parameter1/base-cms-marko-web-identity-x');
 const addOmedaHooksToIdentityXConfig = require('./add-integration-hooks');
+const setPromoSourceCookie = require('./middleware/set-promo-source');
 const stripOlyticsParam = require('./middleware/strip-olytics-param');
 const rapidIdentify = require('./middleware/rapid-identify');
 const rapidIdentifyRouter = require('./routes/rapid-identify');
@@ -14,6 +15,9 @@ module.exports = (app, {
   rapidIdentProductId,
   omedaGraphQLClientProp = '$omedaGraphQLClient',
   omedaRapidIdentifyProp = '$omedaRapidIdentify',
+
+  omedaPromoCodeCookieName = 'oly_promo_src',
+  omedaPromoCodeStripQueryParam = true,
 
   idxConfig,
   idxOmedaRapidIdentifyProp = '$idxOmedaRapidIdentify',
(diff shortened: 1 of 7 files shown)


solocommand:
[Add promo codes to Identity-X configuration](https://trello.com/c/WZOOTUkA/271-add-promo-codes-to-identity-x-configuration)

packages/marko-web-omeda-identity-x/index.js
```diff
@@ -15,6 +16,10 @@ module.exports = (app, {
   omedaGraphQLClientProp = '$omedaGraphQLClient',
   omedaRapidIdentifyProp = '$omedaRapidIdentify',
 
+  omedaPromoCodeCookieName = 'oly_promo_src',
```
zarathustra323: Is the `oly_promo_src` name used by Omeda for something, or was it "made up?" If it isn't an official Omeda param, I'd like to change it to not start with `oly_` otherwise it's hard to tell what is supported by Omeda and what isn't

packages/marko-web-omeda-identity-x/index.js
```diff
@@ -15,6 +16,10 @@ module.exports = (app, {
...
+  omedaPromoCodeStripQueryParam = true,
```
zarathustra323: Is there a specific reason the parameter would need to be stripped? I'm not sure it's necessary, since the param will be processed by our middleware and then set as a cookie

packages/marko-web-omeda-identity-x/integration-hooks/on-login-link-sent.js
```diff
@@ -232,11 +234,14 @@ module.exports = async ({
...
   if (!idxOmedaRapidIdentify) throw new Error(`Unable to find the IdentityX+Omeda rapid identifier on the request using ${idxOmedaRapidIdentifyProp}`);
 
+  const promoCode = req.cookies[omedaPromoCodeCookieName] || defaultPromoCode;
+
   // get omeda customer id (via rapid identity) and load omeda custom field data
   const [{ data }, { encryptedCustomerId }] = await Promise.all([
     identityX.client.query({ query: FIELD_QUERY }),
     idxOmedaRapidIdentify({
       user: user.verified ? user : { id: user.id, email: user.email },
+      ...(promoCode && { promoCode }),
```
zarathustra323, line 249: Can the cookie handling be moved to a lower level (on the `idxOmedaRapidIdentify` call) so that this logic doesn't need to be repeated here and in `middleware/rapid-identify`?

packages/marko-web-omeda-identity-x/middleware/set-promo-source.js
```diff
@@ -0,0 +1,36 @@
...
+ * `oly_promo_src` URL parameter is present.
+ * '
+ * @param {String} omedaPromoCodeCookieName The name of the cookie to interact with
+ * @param {Boolean} stripIncomingPromoCode
+ */
+module.exports = ({
+  omedaPromoCodeCookieName = 'oly_promo_src',
+  stripIncomingPromoCode = true,
+} = {}) => (req, res, next) => {
+  const promoSource = clean(req.cookies[omedaPromoCodeCookieName]);
```
zarathustra323: I'm not sure `clean` is needed here unless, again, this is a cookie officially used by Omeda that also sets string `nulls` and quoted strings

packages/marko-web-omeda-identity-x/middleware/set-promo-source.js
```diff
@@ -0,0 +1,36 @@
...
+
+  const { [omedaPromoCodeCookieName]: promoCode, ...q } = req.query;
+  const incomingPromoCode = clean(promoCode);
+
+  // Only set the promo source cookie if it doesn't already exist
+  if (!promoSource && incomingPromoCode) {
```
zarathustra323: I disagree here... the incoming param should override the current cookie value, otherwise new campaign source attribution will be lost

packages/marko-web-omeda-identity-x/middleware/set-promo-source.js
```diff
@@ -0,0 +1,36 @@
...
+    const options = { maxAge: 60 * 60 * 24 * 365, httpOnly: false };
+    res.cookie(omedaPromoCodeCookieName, `${incomingPromoCode}`, options);
+  }
+
+  // If enabled, strip the URL parameter from the query string
+  if (incomingPromoCode && stripIncomingPromoCode) {
```
zarathustra323: Again, not sure if there's a need to strip this param... would save a lot of unnecessary 302s. The encrypted ID stripping is a little different, and is needed to enforce identity-x user state, but that condition only fires for logged in users also clicking from a newsletters which happens significantly less than promo code params. I don't think stripping is needed

zarathustra323 (changes requested):

packages/marko-web-omeda-identity-x/integration-hooks/on-login-link-sent.js
solocommand: The IdX rapid identify util doesn't have access to the request object (or the additional configs for the cookie name and default value). Rather than pass all that in, I created a util to determine what value to use, and added that in all 3 places. I can switch that around if you think it's warranted...let me know

zarathustra323:
@solocommand can you update the PR description to reflect the changes you made?

solocommand:
> @solocommand can you update the PR description to reflect the changes you made?

Done!

zarathustra323 (approved):

PR #258, opened 2022-03-15
Add ```updateEmailNewsletterProvider``` mutation and resolver.

Shinsina:
Screen Shot 2022-03-15 at 11 09 24 AM

I believe providerId needs to be written as an Int however it is currently writing as a string, I'm guessing this has something to do with the API and how it encodes the update operation as I tried hard casting it as a Number as well as changing the Graph definition to Int instead of String and it still writes it as a string.

EDIT: The previously described behavior is likely due to the Base YAML definition.

### DIFF
--- a/services/graphql-server/src/graphql/definitions/email/newsletter.js
+++ b/services/graphql-server/src/graphql/definitions/email/newsletter.js
@@ -25,6 +25,10 @@ extend type Query {
   )
 }
 
+extend type Mutation {
+  updateEmailNewsletterProvider(input: UpdateEmailNewsletterProviderInput!): EmailNewsletter! @requiresAuth
+}
+
 type EmailNewsletter {
   # fields from platform.model::Product
   id: ObjectID! @projection(localField: "_id") @value(localField: "_id")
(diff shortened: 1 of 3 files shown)


services/graphql-server/src/graphql/definitions/email/newsletter.js
```diff
@@ -150,4 +154,11 @@ input EmailNewsletterSortInput {
   order: SortOrder = desc
 }
 
+input UpdateEmailNewsletterProviderInput {
+  id: ObjectID!
+  type: String
+  providerId: Int
```
solocommand: This should be a `String`, matching the Base [YAML definition](https://github.com/parameter1/base-platform/blob/master/app/config/platform/models/modelspaces/email/deployment.yml#L91) and GraphQL [type definitions](https://github.com/parameter1/base-cms/blob/76dbc184cc9e4c817b8c8d53151f309fc26bf817/services/graphql-server/src/graphql/definitions/email/newsletter.js#L89)

solocommand (changes requested):

solocommand (approved):

PR #263, opened 2022-03-28
Addition of mutation to add/modify company field on content.

Shinsina:
Screen Shot 2022-03-28 at 12 01 13 PM
Screen Shot 2022-03-28 at 12 01 17 PM

### DIFF
--- a/services/graphql-server/src/graphql/definitions/platform/content/index.js
+++ b/services/graphql-server/src/graphql/definitions/platform/content/index.js
@@ -70,6 +70,8 @@ extend type Mutation {
   contentName(input: ContentNameMutationInput!): Content! @requiresAuth
   "Sets the Content User Registration"
   contentUserRegistration(input: ContentUserRegistrationMutationInput!): Content! @requiresAuth
+  "Sets the Content company field"
+  contentCompanyField(input: ContentCompanyFieldMutationInput!): Content! @requiresAuth
 }
 
 enum GateableUserRole {
@@ -783,6 +785,15 @@ input ContentHasWebsiteScheduleInput {
 input QueryMostPopularContentInput {
   siteId: ObjectID
   limit: Int! = 10
+}
+
+input ContentCompanyFieldMutationInput {
+  "The content ID"
+  contentId: Int!
+  "The company ID"
+  companyId: Int!
+  "The mutation to target. The default/non-mutated value will be used when this is null or unset."
+  mutation: ContentMutation
 }
 
 ${interfaces}
--- a/services/graphql-server/src/graphql/definitions/platform/content/index.js
+++ b/services/graphql-server/src/graphql/definitions/platform/content/index.js
@@ -789,7 +789,7 @@ input QueryMostPopularContentInput {
 
 input ContentCompanyFieldMutationInput {
   "The content ID"
-  contentId: Int!
+  id: Int!
   "The company ID"
   companyId: Int!
   "The mutation to target. The default/non-mutated value will be used when this is null or unset."
--- a/services/graphql-server/src/graphql/definitions/platform/content/index.js
+++ b/services/graphql-server/src/graphql/definitions/platform/content/index.js
@@ -792,8 +792,6 @@ input ContentCompanyFieldMutationInput {
   id: Int!
   "The company ID"
   companyId: Int!
-  "The mutation to target. The default/non-mutated value will be used when this is null or unset."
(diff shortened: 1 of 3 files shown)


services/graphql-server/src/graphql/definitions/platform/content/index.js
```diff
@@ -785,6 +787,15 @@ input QueryMostPopularContentInput {
   limit: Int! = 10
 }
 
+input ContentCompanyFieldMutationInput {
+  "The content ID"
+  id: Int!
+  "The company ID"
+  companyId: Int!
+  "The mutation to target. The default/non-mutated value will be used when this is null or unset."
+  mutation: ContentMutation
```
solocommand: `company` isn't a mutation-enabled, remove this input field. The mutation referred to here is the Base Platform mutation (allowing a different version of a field based on expected target rendition, such as website/newsletter/print), not the GraphQL mutation field.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -1841,6 +1841,18 @@ module.exports = {
...
     },
 
+    /**
+     *
+     */
+    contentCompanyField: updateContentMutationHandler({
+      buildPayload: (input) => {
+        const value = input.companyId;
+        if (!value && !input.mutation) throw new UserInputError('The default mutation of the content name cannot be empty.');
+        const field = input.mutation ? `company${input.mutation}` : 'company';
```
solocommand: Remove mutation logic here. Since this is a reference field and not a flat value, I believe this needs to be under Links. Check the product/image relationship resolvers for examples.

solocommand (changes requested):
Remove platform mutation references. Boot and test against p1 sandbox please!

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -1841,6 +1841,25 @@ module.exports = {
...
+     *
+     */
+    contentCompanyField: async (_, { input }, { base4rest, basedb }, info) => {
+      validateRest(base4rest);
+      const company = 'platform/content/company';
+      const { id, companyId } = input;
+      const content = await basedb.strictFindById('platform.Content', id, { projection: { type: 1 } });
+      const model = `platform/content/${dasherize(content.type)}`;
+      const body = new Base4RestPayload({ type: model });
+      if (companyId) body.setLink('company', { id: companyId, type: company });
```
solocommand: `company` should be nullable -- the `setLink` and `updateOne` should always happen regardless of a value. Update the definition for the input to allow the company to be nulled out and add a note about nullability? to the `companyId` field -- something like "If not present, the company relationship will be removed"

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -1841,6 +1841,25 @@ module.exports = {
...
+      if (content) {
+        body.set('id', id);
+        await base4rest.updateOne({ model, id, body });
+      }
```
solocommand: `id` is not nullable, and `strictFindById` will error before reaching this line if a bad ID is passed. No need for the if check.

solocommand (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -1841,6 +1841,24 @@ module.exports = {
...
+     */
+    contentCompanyField: async (_, { input }, { base4rest, basedb }, info) => {
+      validateRest(base4rest);
+      const company = 'platform/content/company';
+      const { id, companyId } = input;
+      const content = await basedb.strictFindById('platform.Content', id, { projection: { type: 1 } });
+      const model = `platform/content/${dasherize(content.type)}`;
+      const body = new Base4RestPayload({ type: model });
+      if (companyId) body.setLink('company', { id: companyId, type: company });
+      else body.unsetLink('company');
```
solocommand: Not sure why the linter is missing this, but fix your if/else clause: https://github.com/airbnb/javascript#blocks

solocommand (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
solocommand: Bonus points for fixing the linter :)

solocommand (approved):

PR #270, opened 2022-04-05
Add IdentityX data layer events

B77Mills:
Add push events to the dataLayerIdentityX for login, profile view & profile submit.  This also allows for passing custom event names into the given events.


packages/marko-web-identity-x/browser/profile.vue
```diff
@@ -234,6 +238,18 @@ export default {
       type: Array,
       default: () => [],
     },
+    eventName: {
+      type: String,
+      default: 'profile',
+    },
```
solocommand: `eventName` isn't used within this component, only `viewEventName` and `submitEventName`.

packages/marko-web-identity-x/utils/gtm-events.js
```diff
@@ -0,0 +1,18 @@
...
+  if (dataLayer) dataLayer.push({ event, email });
+};
+
+export const profileView = ({ event = 'profile_view' }) => {
+  console.log('gtm-event-profile-view: ', event);
+  const dataLayer = window.dataLayerIdentityX;
+  if (dataLayer) dataLayer.push({ event });
+};
+
+export const profileSubmit = ({ event = 'update_profile', email }) => {
```
solocommand: You made the default value in the Vue component `profile_submit`, but the default value here is `update_profile`. Which is correct?

packages/marko-web-identity-x/utils/gtm-events.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-undef */
+export const loginSubmit = ({ event = 'login', email }) => {
+  console.log('gtm-event-login-submit: ', event);
```
solocommand: Remove logs

packages/marko-web-identity-x/components/form-profile.marko
```diff
@@ -12,6 +12,9 @@ $ const { identityX } = req;
       hiddenFields: identityX.config.getHiddenFields(),
       callToAction: input.callToAction,
       reloadPageOnSubmit: input.reloadPageOnSubmit,
+      viewEventName: input.viewEventName,
+      submitEventName: input.submitEventName,
+      reloadPageOnSubmit: input.reloadPageOnSubmit,
```
solocommand: `reloadPageOnSubmit` is now duplicated in the component props

solocommand (changes requested):
A few changes. Be careful of implementation on the GTM side -- the email address **is** considered PII. Ultimately it's up to the implementor to ensure they are in compliance. Is there a specific request for the email address to be sent, or would the IdX user id suffice?

packages/marko-web-identity-x/browser/profile.vue
B77Mills: I removed the eventName prop that was never used instead handle with the viewEventName & submitEventName props

solocommand (dismissed):
@B77Mills looks good, but there's a conflict in the marko component definitions

B77Mills:
@solocommand I removed the passing of email address on login completely and added the passing of userId instead of email on profile is submit.

solocommand:
@brandonbk @B77Mills updated to include event dispatching. Notable changes:
- pass `eventLabel` instead of `eventName` to adjust the reported property
- Comment post view does not emit an event (noisy)
- There is still a race condition where events on pages with client-side redirects (authenticate, logout) may not consistently report

solocommand (approved):

PR #276, opened 2022-04-11
Add original value length check to truncateTeaser

B77Mills:
Add the check for the length of the value being longer that maxLength when the html is stripped and spaces are removed. If not return the original value.

### DIFF
--- a/services/graphql-server/src/graphql/utils/content-teaser.js
+++ b/services/graphql-server/src/graphql/utils/content-teaser.js
@@ -10,9 +10,11 @@ const getTeaser = (mutation, content) => {
 };
 
 const truncateTeaser = (value, { maxLength, truncatedSuffix }) => {
-  if (!maxLength || stripHtml(value).replace(' ', '').length <= maxLength) return value;
+  if (!maxLength) return value;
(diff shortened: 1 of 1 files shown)


jwade1327 (dismissed):

services/graphql-server/src/graphql/utils/content-teaser.js
```diff
@@ -10,7 +10,7 @@ const getTeaser = (mutation, content) => {
 };
 
 const truncateTeaser = (value, { maxLength, truncatedSuffix }) => {
-  if (!maxLength) return value;
+  if (!maxLength || stripHtml(value).replace(' ', '').length <= maxLength) return value;
   // @todo This strips HTML from truncated teasers. We could attempt to preserve it?
   const stripped = stripHtml(value);
```
solocommand: No need to run the strip twice

```suggestion
  if (!maxLength) return value;
  // @todo This strips HTML from truncated teasers. We could attempt to preserve it?
  const stripped = stripHtml(value);
  if (stripped.replace(' ', '').length <= maxLength) return value;
```

solocommand (changes requested):

solocommand (approved):

PR #279, opened 2022-04-19
IdentityX form classing & display logic adjustments

B77Mills:
Make the col-md-6 class dynamic based on it sibling field in each row.

Add hiddenFields prop to the authenticate form components.

Update Address block visibility checks
 - add generic address-block class for css targeting purposes
 - add or for postalCode visibility check on line 2 of the address block
 - update computed class logic for line two to support full 1/2 or 1/3(col-me-12||6||4…
 - update v-if on postalCode to check itself instead of regionCode


B77Mills:
[Metered Paywall on Overdrive](https://trello.com/c/scDTaRw2/111-metered-paywall-on-overdrive)

solocommand (dismissed):

PR #281, opened 2022-04-20
Set US as default country value

brandonbk:


solocommand:
Closed in favor of #282

PR #284, opened 2022-04-21
Addition of ```exists``` parameter to ContentCustomAttributeQueryInput. 

Shinsina:
Edited now using ```exists``` input parameter.

Screen Shot 2022-04-21 at 3 35 05 PM
Screen Shot 2022-04-21 at 3 35 13 PM
Screen Shot 2022-04-21 at 3 35 41 PM
Screen Shot 2022-04-21 at 3 35 54 PM
Screen Shot 2022-04-21 at 3 36 03 PM
Screen Shot 2022-04-21 at 3 34 51 PM

### DIFF
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -918,8 +918,17 @@ module.exports = {
         excludeContentTypes,
       });
 
-      customAttributes.forEach(({ key, value, isJSON }) => {
-        query.$and.push({ [`customAttributes.${key}`]: isJSON ? JSON.parse(value) : value });
+      customAttributes.forEach(({ key, value, exists }) => {
+        if (!exists) {
+          query.$and.push({
+            $or: [
+              { [`customAttributes.${key}`]: { $exists: false } },
+              { [`customAttributes.${key}`]: { $in: ['', null] } },
+            ],
+          });
+        } else {
+          query.$and.push({ [`customAttributes.${key}`]: value });
+        }
       });
 
       const siteId = input.siteId || site.id();
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -919,17 +919,15 @@ module.exports = {
       });
 
       customAttributes.forEach(({ key, value, exists }) => {
-        const valueOrExists = Boolean(value || typeof exists !== 'undefined');
-        if (!valueOrExists) {
-          throw new UserInputError('Either value or exists must be defined');
-        } else {
-          query.$and.push({
-            [`customAttributes.${key}`]: {
-              ...((value && exists !== false) ? { $nin: ['', null], $exists: true } : { $in: ['', null] }),
-            },
-          });
-          query.$and.push({ [`customAttributes.${key}`]: value });
-        }
+        const valueOrExists = Boolean(value || typeof exists === 'boolean');
+        if (!valueOrExists) throw new UserInputError('Value or exists must be defined');
+        query.$and.push({
+          [`customAttributes.${key}`]: {
+            ...(value && { $eq: value }),
+            ...(exists === false && { $in: ['', null] }),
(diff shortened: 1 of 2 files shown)


Shinsina:
[GraphQL guidance: how to filter a query to find content items with page number but no Digital Edition URL](https://trello.com/c/hOZVNkjO/297-graphql-guidance-how-to-filter-a-query-to-find-content-items-with-page-number-but-no-digital-edition-url)

zarathustra323:
Adding query support to the value input opens the door for all sorts of potential issues.

zarathustra323:
Since it isn't restricted, any type of query condition could be executed potentially causing the API to crash the DB

solocommand:
@zarathustra323 agreed. @Shinsina, change this to add a query input explicitly for what you need. Something like
```diff
  key: String!
  value: String!
+ exists: Boolean
```
And then handle that explicitly within the resolver.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -918,8 +918,17 @@ module.exports = {
...
+        if (!exists) {
+          query.$and.push({
+            $or: [
+              { [`customAttributes.${key}`]: { $exists: false } },
+              { [`customAttributes.${key}`]: { $in: ['', null] } },
+            ],
+          });
+        } else {
+          query.$and.push({ [`customAttributes.${key}`]: value });
+        }
```
```suggestion
      customAttributes.forEach(({ key, value, exists }) => {
        // error if both value and exists are null
        if (exists !== null) {
        query.$and.push({
          [`customAttributes.${key}`]: {
            $exists: Boolean(exists),
            ...(exists ? { $nin: ['', null] } : { $in: ['', null] }),
          },
        };
        }
        if (!exists) {
          query.$and.push({
            $or: [
              { [`customAttributes.${key}`]: { $exists: false } },
              { [`customAttributes.${key}`]: { $in: ['', null] } },
            ],
          });
        } else {
          query.$and.push({ [`customAttributes.${key}`]: value });
        }
```

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -918,8 +918,18 @@ module.exports = {
         excludeContentTypes,
       });
 
-      customAttributes.forEach(({ key, value }) => {
-        query.$and.push({ [`customAttributes.${key}`]: value });
+      customAttributes.forEach(({ key, value, exists }) => {
+        const valueOrExists = Boolean(value || typeof exists !== 'undefined');
```
solocommand: Technically, this should check that `typeof exists === 'boolean'`. We want to error if there is no value and exists is nullish.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -918,8 +918,18 @@ module.exports = {
...
+        if (!valueOrExists) {
+          throw new UserInputError('Either value or exists must be defined');
+        } else {
```
solocommand: Use condensed if here if it will fit; `if(!validInput) throw new UserInputError(...);`. No need to wrap the rest in an else since the throw will abort execution of the rest.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -918,8 +918,18 @@ module.exports = {
...
+          query.$and.push({
+            [`customAttributes.${key}`]: {
+              ...((value && exists !== false) ? { $nin: ['', null], $exists: true } : { $in: ['', null] }),
+            },
+          });
+          query.$and.push({ [`customAttributes.${key}`]: value });
```
solocommand: Since we're changing it up with the exists operators, use `$eq` for value to keep it consistent. Something like this:
```js
[cattr.${key}]: {
  ...(value && { $eq: value }),
  ...(exists === false && { $exists: false, $nin, $in, ... }),
  ...(exists === true && { $exists: true, ...  }),
```

solocommand (changes requested):

solocommand (approved):

PR #285, opened 2022-04-21
Resolve magazine schedules on Content interface.

Shinsina:
Screen Shot 2022-04-21 at 1 46 08 PM

### DIFF
--- a/services/graphql-server/src/graphql/definitions/platform/content/index.js
+++ b/services/graphql-server/src/graphql/definitions/platform/content/index.js
@@ -250,21 +250,6 @@ type ContentWebsiteSchedule {
   endDate(input: FormatDate = {}): String @momentFormat(localField: "end")
 }
 
-type ContentMagazineSchedule {
-  # fields from magazine.model::Schedule
-  id: ObjectID! @projection(localField: "_id") @value(localField: "_id")
-  publication(input: MagazineSchedulePublicationInput = {}): MagazinePublication! @projection(localField: "product") @refOne(
-    loader: "platformProduct",
-    localField: "product"
-    criteria: "magazinePublication",
-  )
-  issue(input: MagazineScheduleIssueInput = {}): MagazineIssue! @projection @refOne(loader: "magazineIssue")
-  section(input: MagazineScheduleSectionInput = {}): MagazineSection! @projection @refOne(loader: "magazineSection")
-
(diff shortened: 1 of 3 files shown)


Shinsina:
[GraphQL guidance: how to filter a query to find content items with page number but no Digital Edition URL](https://trello.com/c/hOZVNkjO/297-graphql-guidance-how-to-filter-a-query-to-find-content-items-with-page-number-but-no-digital-edition-url)

services/graphql-server/src/graphql/definitions/platform/content/index.js
```diff
@@ -250,6 +250,21 @@ type ContentWebsiteSchedule {
...
+    localField: "product"
+    criteria: "magazinePublication",
+  )
+  issue(input: MagazineScheduleIssueInput = {}): MagazineIssue! @projection @refOne(loader: "magazineIssue")
+  section(input: MagazineScheduleSectionInput = {}): MagazineSection! @projection @refOne(loader: "magazineSection")
+
+  # fields from trait.platform::StatusEnabled
+  status: Int @projection
+}
+
```
solocommand: Pretty sure you don't need to define this; `ContentWebsiteSchedule` was used instead of `WebsiteSchedule` because the source data had different local field mappings (coming from `sectionQuery` instead of the website schedule directly).

You should be able to use `MagazineSchedule` and not define this extra type here

solocommand (changes requested):
@

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -444,6 +444,12 @@ module.exports = {
       return isArray(sectionQuery) ? sectionQuery : [];
     },
 
+
+    magazineSchedules: async (content, _, { basedb }) => {
+      const contentMagazineSchedules = await basedb.find('magazine.Schedule', { 'content.$id': content._id });
+      return contentMagazineSchedules;
+    },
```
```suggestion
    magazineSchedules: ({ _id }, _, { basedb }) => basedb.find('magazine.Schedule', { 'content.$id': _id }),
```

Just to clean this up a bit: There's no need to define a variable just to return it, and no need to make this an async function when only awaiting one thing. Just return the promise directly

solocommand (changes requested):

solocommand (approved):

PR #289, opened 2022-04-25
Update IdentityX event emission

solocommand:
Now uses injected `EventBus` directly, moves GTM/dataLayer code to GTM package.

Closes #288 

Requires parameter1/identity-x#11


packages/marko-web-identity-x/browser/service.js
```diff
@@ -0,0 +1,27 @@
...
+};
+const { warn } = console;
+
+export default class IdentityX {
+  context = {};
+
+  constructor() {
+    try {
+      // Read cookie
+      const cookie = getCookie('__idx_ctx');
```
zarathustra323:
```js
const parse = (json) => {
  try {
    return JSON.parse(json);
  } catch (e) {
    return {};
  }
}
```

packages/marko-web-identity-x/browser/service.js
```diff
@@ -0,0 +1,27 @@
+// It's 2022, why is there not a better way to do this?
+const getCookie = (key) => {
+  const cookies = document.cookie.split(/; */).reduce((obj, cookie) => {
```
zarathustra323: Use string `;` in the split with trim... do not add to object if trimmed value is falsey

packages/marko-web-identity-x/browser/service.js
```diff
@@ -0,0 +1,27 @@
...
+    const [name, value] = cookie.split('=').map(decodeURIComponent);
```
zarathustra323: `.map` should decode and trim

zarathustra323 (changes requested):

packages/marko-web-gtm/browser/index.js
```diff
@@ -11,4 +11,47 @@ export default (Browser) => {
...
+    'identity-x-authenticate-errored',
+    'identity-x-comment-post-errored',
+    'identity-x-comment-report-errored',
+    'identity-x-comment-stream-errored',
+    'identity-x-login-errored',
+    'identity-x-logout-errored',
+    'identity-x-profile-errored',
+  ].forEach((event) => {
+    EventBus.$on(event, (args) => {
+      if (window.IdentityX) {
```
zarathustra323: Would be a little cleaner as `if (!window.IdentityX) return;` then the data layer push

packages/marko-web-identity-x/browser/service.js
```diff
@@ -0,0 +1,28 @@
+// It's 2022, why is there not a better way to do this?
+const getCookie = (key) => {
+  const cookies = document.cookie.split(';').reduce((obj, cookie) => {
+    const [name, value] = cookie.split('=').map(v => decodeURIComponent(v.trim()));
```
zarathustra323:
```js
const trimmed = cookie.trim();
if (!trimmed) return obj;

File: packages/marko-web-identity-x/browser/service.js [hunk 6]
```diff
@@ -0,0 +1,28 @@
...
+const parseContext = () => {
+  try {
+    return JSON.parse(getCookie('__idx_ctx'));
+  } catch (e) {
+    return {};
+  }
+};
+
+export default class IdentityX {
+  context = {};
```
zarathustra323: Get rid of me :)

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #294, opened 2022-05-02
Write-in value support for IdentityX custom selects

solocommand:
Adds support for displaying and updating write-in values on the user profile.

with write-in value
image

with normal value
image

autofocus on empty value
image


solocommand:
[Support write-in values on select fields](https://trello.com/c/GB789yV5/3-support-write-in-values-on-select-fields)

B77Mills (dismissed):

PR #300, opened 2022-05-11
Move i18n prop destructure to correct location.

Shinsina:
This was causing subsequent pages when paginating on taxonomy to error out and display nothing as the i18n function is not defined on input but rather app.locals (out.global).

Screen Shot 2022-05-11 at 9 40 30 AM


Shinsina:
[Investigate pagination issue](https://trello.com/c/18IDNYQU/128-investigate-pagination-issue)

brandonbk (approved):

PR #313, opened 2022-05-27
Strip oly_enc_id from query param URL.

Shinsina:
This only happens on production in non-incognito windows, I can’t figure out a way of testing this without just putting it on production.

I seemed to recall something about this being done deliberately, however on production in an incognito window the query parameter is properly stripped so I don't entirely know what’s going on here.

(Screenshot from production where it is an issue)
Screen Shot 2022-05-27 at 12 19 27 PM

### DIFF
--- a/packages/marko-web-p1-fii/browser/omeda-alchemer-form.vue
+++ b/packages/marko-web-p1-fii/browser/omeda-alchemer-form.vue
@@ -89,7 +89,13 @@ export default {
         });
         const json = await res.json();
         if (!res.ok) throw new Error(json.message);
-        return json.data.url;
+        const rootUrl = new URL(json.data.url);
+        const urlParams = new URLSearchParams(rootUrl.search);
+        const paramUrl = new URL(urlParams.get('url'));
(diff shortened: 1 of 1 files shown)


Shinsina:
[Event Coverage Category Mapping](https://trello.com/c/r0YG84TR/293-event-coverage-category-mapping)

Shinsina:
[Encrypted ID is being passed to Alchemer form](https://trello.com/c/ci2yjMDA/300-encrypted-id-is-being-passed-to-alchemer-form)

Shinsina:
This appears to be due to the query parameters being passed without regard as to what they are see: https://github.com/parameter1/form-identity-integrations/blob/main/api/src/actions/form/wufoo/prepop-with-omeda.js#L217 and https://github.com/parameter1/form-identity-integrations/blob/main/api/src/actions/form/alchemer/prepop-with-omeda.js#L158. Correction will need to be made there to account for (potentially optionally) removing URL query parameters from the page requesting the given form.

PR #316, opened 2022-05-31
Use native browser lazyloading

brandonbk:


zarathustra323 (commented):
Is the lazysizes lib used anymore? If not the JS should also be removed

brandonbk:
@zarathustra323 - yes, it is still being used by some Vue components within base-cms as well as customer repos.  Those will be handled with a separate PR, and once tested we will circle back and remove lazysizes.

B77Mills (approved):

PR #322, opened 2022-06-02
Add <picture> support

brandonbk:


packages/marko-web-theme-monorail/components/blocks/primary-image.marko
```diff
@@ -62,14 +80,19 @@ $ if (image.id) imageAttrs["data-image-id"] = image.id;
...
-        src=src
-        srcset=srcset
-        alt=image.alt
-        class=`${blockName}__image`
-        attrs=imageAttrs
-        lazyload=lazyload
-      />
+      <marko-web-picture>
+        <for|s| of=sources>
+          <@source ...s />
```
solocommand: This should be explicitly defined, not spread, in order to pass argument validation. If the required arguments for the picture source component change in the future, this should now error because the wrong props are being sent.
```suggestion
          <@source srcset=s.srcset media=s.media />
```

packages/marko-web/components/element/picture.marko
```diff
@@ -0,0 +1,25 @@
+import { getAsArray } from "@parameter1/base-cms-object-path";
+
+$ const hasImage = Boolean(input.image.src);
+
+<if(hasImage)>
+  <if(input.link)>
+    <marko-core-link ...input.link>
```
solocommand: Same here -- these attributes should be explicitly defined. Spreading them will bypass validation -- either remove the spread and pass the props explicitly, or remove the validation rules from the Marko component definition.

packages/marko-web/components/element/picture.marko
```diff
@@ -0,0 +1,25 @@
...
+    $ const sources = getAsArray(input, "sources");
+    <picture>
+      <for|s| of=sources>
+        $ const srcset = s.srcset.join(", ");
+        <source
+          srcset=srcset
+          media=s.media
+        >
+      </for>
+      <image ...input.image />
```
solocommand: Here too

packages/marko-web/components/element/picture.marko
```diff
@@ -0,0 +1,25 @@
+import { getAsArray } from "@parameter1/base-cms-object-path";
+
+$ const hasImage = Boolean(input.image.src);
+
+<if(hasImage)>
+  <if(input.link)>
+    <marko-core-link ...input.link>
+      $ input.link = false;
```
solocommand: This should be unnecessary if defining the props manually. If not, use spreading to separate the attributes, rather than overwriting them.
```marko
$ const { link, ...props } = input;
<if(link)>
  <marko-web-picture ...props />
```

solocommand (changes requested):

solocommand (approved):

PR #325, opened 2022-06-07
Allow zero values in `x1` and `y1` crop dimensions

Shinsina:
From Trello card description: 

"Currently the logic found within https://github.com/parameter1/base-cms/blob/master/packages/image/src/crop-rectangle.js results in if a crop rectangle has coordinates of ```x = 0``` and ```y = 0``` it determines that an image has not been cropped, by default x and y are set to x1 and y1 meaning that if x1 and y1 are both 0 it determines the image has not been cropped regardless of it cropDimensions are set on the image or not.

x1 and y1 of 0 and 0 respectively are perfectly valid so this needs to be corrected to perform some other sort of logic in order to determine whether or not an image has been cropped."

Corrected Item:

- GraphQL:
Screen Shot 2022-06-07 at 12 10 41 PM

- MongoDB Asset Collection:
Screen Shot 2022-06-07 at 12 02 36 PM

- Logging GraphQL Server Response:
Screen Shot 2022-06-07 at 12 01 51 PM

Unaffected Item:

- GraphQL:
Screen Shot 2022-06-07 at 12 10 48 PM

- MongoDB Asset Collection:
Screen Shot 2022-06-07 at 12 17 26 PM

### DIFF
--- a/packages/image/src/crop-rectangle.js
+++ b/packages/image/src/crop-rectangle.js
@@ -4,19 +4,21 @@ class CropRectangle {
     y,
     width,
     height,
+    cropped,
   } = {}) {
     this.x = x;
     this.y = y;
     this.width = width;
     this.height = height;
+    this.cropped = cropped;
   }
 
   isCropped() {
-    return !this.notCropped();
+    return this.cropped;
   }
 
   notCropped() {
-    return this.x === 0 && this.y === 0;
+    return !this.cropped;
   }
 
   toString() {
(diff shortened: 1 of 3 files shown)


Shinsina:
[Correct logic for determining whether or not an image has been cropped](https://trello.com/c/rXsPV4sE/101-correct-logic-for-determining-whether-or-not-an-image-has-been-cropped)

B77Mills (dismissed):

zarathustra323 (approved):

PR #329, opened 2022-06-08
Compute aria-label for menu-toggle-button.

Shinsina:
Currently the ```aria-label``` for the content meter is not reflective as to what visibly appears on the page thus it errors, this corrects it to use the same value in both placements when and where applicable.

Dev-Menu-Toggle
Prod-Menu-Toggle


Shinsina:
[Accessibility improvements](https://trello.com/c/7OYysXzp/130-accessibility-improvements)

brandonbk (approved):

PR #341, opened 2022-06-13
Remove text-align: center from embedded images

brandonbk:


brandonbk:
[Embedded image caption text should not be center aligned](https://trello.com/c/nxToYZEm/139-embedded-image-caption-text-should-not-be-center-aligned)

PR #344, opened 2022-06-14
Adjust inline newsletter box

B77Mills:
Add support for the inline newsletter signup box to be called on section pages & to span full width.  This changes moves the background color to the wrapping inline element form with the form having a max width of 652px & margin left/right auto. This will force the form to appear aligned centered.  
Screen Shot 2022-06-14 at 10 38 26 AM
Screen Shot 2022-06-14 at 10 38 35 AM
Screen Shot 2022-06-14 at 10 38 45 AM
Screen Shot 2022-06-14 at 10 38 57 AM

### DIFF
--- a/packages/marko-web-theme-monorail/scss/components/_site-newsletter-menu.scss
+++ b/packages/marko-web-theme-monorail/scss/components/_site-newsletter-menu.scss
@@ -170,7 +170,7 @@
     font-size: 18px;
     font-weight: 700;
     line-height: 1;
-    color: #fff;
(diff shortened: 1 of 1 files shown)


Shinsina (approved):

B77Mills:
[Add full-width newsletter signup block to home page](https://trello.com/c/YOeNy42a/360-add-full-width-newsletter-signup-block-to-home-page)

zarathustra323 (approved):

PR #348, opened 2022-06-15
Add with-native-x-section support to list flow

B77Mills:
- Add new with-native-x-section prop to list flow
 — New prop is used in conjunction with withSection, node.id & nxNode.id to determine if it should display the section tag(Sponsored) even when withSection is false.
 - Add pass through support from the latest-content-list block

Screen Shot 2022-06-15 at 2 35 08 PM


B77Mills:
[Move 'Related Stories' block from below article to right rail on all content pages (add NativeX ad slot to block)](https://trello.com/c/tXHjJCab/255-move-related-stories-block-from-below-article-to-right-rail-on-all-content-pages-add-nativex-ad-slot-to-block)

zarathustra323 (approved):

PR #357, opened 2022-06-27
Adjust when `strip-olytics-params.js` middleware should redirect & clean url.

B77Mills:
When no IdentityX context  is set, but there is an incoming oly_enc_id it now set the olytics cookie without forcing redirect.  It will now only redirect when there is an incoming Id conflict.  Meaning there is already an idX user context present that doesn't match the incoming request oly_enc_id.  Basically remove it and reload the page with the already set idX user context.

### DIFF
--- a/packages/marko-web-omeda-identity-x/middleware/strip-olytics-param.js
+++ b/packages/marko-web-omeda-identity-x/middleware/strip-olytics-param.js
@@ -8,16 +8,12 @@ module.exports = () => (req, res, next) => {
   const { oly_enc_id: id, ...q } = req.query;
   const incomingEncId = olyticsCookie.clean(id);
 
-  // No IdentityX context, set the cookie and move on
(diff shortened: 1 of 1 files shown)


Shinsina:
[Encrypted ID is being passed to Alchemer form](https://trello.com/c/ci2yjMDA/300-encrypted-id-is-being-passed-to-alchemer-form)

solocommand (changes requested):
Re-implement this as a new middleware in the Omeda package that reads/returns/sets the context. By default it should read from the request cookies. In this middleware (in Omeda+IdX), it should call `omedaCtx.setEncryptedId(...)`, which should set the cookie to the response.

In the final case (mismatch), it should redirect without setting the cookie, as it does currently. This is to explictly bypass the Omeda client-side parameter handling.

solocommand (approved):

PR #370, opened 2022-07-08
Set default value & spread props only when set

B77Mills:
This was trying to spread an undefined prop instead set a default of an empty object.


packages/marko-web-theme-monorail/components/site-footer.marko
```diff
@@ -4,6 +4,7 @@ import { asObject } from "@parameter1/base-cms-utils";
 
 $ const { config, site, i18n } = out.global;
 $ const newsletterSignupConfigName = defaultValue(input.newsletterSignupConfigName, "signupFooter");
+$ const newsletterProps = defaultValue(input.newsletterBlockProps, {});
```
zarathustra323: Since this is expects an object I'd use getAsObject instead of defaultValue

zarathustra323 (dismissed):

solocommand (approved):

PR #383, opened 2022-07-25
Adjust comment stream css to account for comment count

B77Mills:
Move max-height and hiding of overflow done from the wrapper to the actual comment stream.  This allows use to account for the where there are no comments or a single comment displaying the whole stream.

By default now the $comment-start-height ~ max-height will be applied stream instead of the wrapper(stream + toggle button).  It will also not set it when there are no comments or only a single comment.

Also moved the show comments transition when the toggle button is pressed.

### Multi Comments:
Screen Shot 2022-07-25 at 10 53 59 AM

### Single Comment:
Screen Shot 2022-07-25 at 10 53 41 AM

### No Comments
Screen Shot 2022-07-25 at 11 00 45 AM
:


B77Mills:
[AQUA Comment](https://trello.com/c/Z8BIf5bf/313-aqua-comment)

solocommand (approved):

PR #386, opened 2022-08-01
Allow section renaming in list blocks, Monorail

Shinsina:
podcasts-section-homepage

(The podcasts section shown here is actually called AW Podcasts for clarify from the editorial/management side so the name needs to be adjusted on site to only show "Podcasts")


Shinsina:
[Subsection for AW Podcasts](https://trello.com/c/HYZwLgVX/356-subsection-for-aw-podcasts)

brandonbk (approved):

PR #387, opened 2022-08-02
Update attribute handling in `updateEmailNewsletterProvider` mutation

Shinsina:
code-side
graph-side

### DIFF
--- a/services/graphql-server/src/graphql/definitions/email/newsletter.js
+++ b/services/graphql-server/src/graphql/definitions/email/newsletter.js
@@ -161,7 +161,7 @@ input UpdateEmailNewsletterProviderInput {
   type: String
   "The provider id for the newsletter product from an email service provider"
   providerId: String
-  "Setting an key in attributes value to null, will subsequently remove that key on the respective Newsletter Product's attributes,"
+  "The key-values to append to the newsletter provider attributes. Send `null` value to unset a key."
   attributes: JSON
 }
 
--- a/services/graphql-server/src/graphql/definitions/email/newsletter.js
+++ b/services/graphql-server/src/graphql/definitions/email/newsletter.js
@@ -155,9 +155,13 @@ input EmailNewsletterSortInput {
 }
 
 input UpdateEmailNewsletterProviderInput {
+  "The id of the newsletter product to update the provider information for"
   id: ObjectID!
+  "The provider type for the email service provider (typically set to omeda)"
   type: String
+  "The provider id for the newsletter product from an email service provider"
   providerId: String
+  "Setting an key in attributes value to null, will subsequently remove that key on the respective Newsletter Product's attributes,"
   attributes: JSON
 }
 
--- a/services/graphql-server/src/graphql/resolvers/email/newsletter.js
+++ b/services/graphql-server/src/graphql/resolvers/email/newsletter.js
@@ -9,8 +9,9 @@ module.exports = {
    */
   Mutation: {
     /**
-      * Setting an key in attributes value to null,
-      * will subsequently remove that key on the respective Newsletter Product,
+    * @description Update email newsletter provider fields.
+    * Setting an key in attributes value to null,
+    * will subsequently remove that key on the respective Newsletter Product's attributes,
     */
     updateEmailNewsletterProvider: async (_, { input }, { base4rest, basedb }) => {
       validateRest(base4rest);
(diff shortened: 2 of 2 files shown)


services/graphql-server/src/graphql/resolvers/email/newsletter.js
```diff
@@ -17,10 +18,26 @@ module.exports = {
         attributes,
       } = input;
 
+      const existingNewsletterProduct = await basedb.findOne('platform.Product', { _id: id }, { projection: { provider: 1 } });
```
solocommand: Just call this `product`, in this context there's no need to distinguish it from other product types, or a newly created one. Also, use the `strictFindById` method here.

services/graphql-server/src/graphql/resolvers/email/newsletter.js
```diff
@@ -17,10 +18,26 @@ module.exports = {
...
+      const existingAttributes = get(existingNewsletterProduct, 'provider.attributes');
```
solocommand: Use `getAsObject` rather than `get`. In this case, if you retrieve an item that doesn't have the attributes object set, you will hard error on L25 & L28.

services/graphql-server/src/graphql/resolvers/email/newsletter.js
```diff
@@ -17,10 +18,26 @@ module.exports = {
...
+          mergedAttributes[attribute] = attributes[attribute];
+        } else {
+          mergedAttributes[attribute] = existingAttributes[attribute];
+        }
+      });
+      Object.keys(attributes).forEach((attribute) => {
+        if (!mergedAttributes[attribute] && typeof attributes[attribute] !== 'undefined') {
+          mergedAttributes[attribute] = attributes[attribute];
+        }
+      });
```
solocommand: Use a reducer, `deepmerge`, or `lodash`. You don't need to iterate over the keys of both objects, start with the existing attributes and update/unset as necessary based on the incoming data (if any).

A simpler approach:
```js
const attrs = { ...attributes, ...newAttrs };
Object.entries(newAttrs).forEach(([k, v]) => {
  if (null === v) delete attrs[k];
});
```

solocommand (changes requested):

services/graphql-server/src/graphql/resolvers/email/newsletter.js
```diff
@@ -7,6 +8,10 @@ module.exports = {
    *
    */
   Mutation: {
+    /**
+      * Setting an key in attributes value to null,
+      * will subsequently remove that key on the respective Newsletter Product,
+    */
```
solocommand: Sorry if I was unclear -- the comment should be added to the GraphQL field definition. This is fine to keep, but fix the indentation and include a description of the mutation itself -- updates email newsletter provider fields, etc.

For the GraphQL doc, something [like this](https://github.com/parameter1/mindful-cms/blob/main/graphql/managed/src/definitions/email-block.js#L64), but specific to the mutation and input logic. The doc line should be added above the `attributes` field in the type definition.

solocommand (changes requested):

services/graphql-server/src/graphql/resolvers/email/newsletter.js
Shinsina: All good! I wasn't sure which "GraphQL definition" you were talking about so I thought it made sense to put it on the mutation resolver, I've added a description to the mutation resolver as well as to the fields on the input definition. Do let me know if anything needs to be changed, improved or added anywhere!

services/graphql-server/src/graphql/definitions/email/newsletter.js
```diff
@@ -155,9 +155,13 @@ input EmailNewsletterSortInput {
 }
 
 input UpdateEmailNewsletterProviderInput {
+  "The id of the newsletter product to update the provider information for"
   id: ObjectID!
+  "The provider type for the email service provider (typically set to omeda)"
   type: String
+  "The provider id for the newsletter product from an email service provider"
   providerId: String
+  "Setting an key in attributes value to null, will subsequently remove that key on the respective Newsletter Product's attributes,"
```
solocommand: I'd prefer this still be descriptive of the field itself. Something like this?
```suggestion
  "The key-values to append to the newsletter provider attributes. Send `null` value to unset a key."
```

solocommand (changes requested):

PR #395, opened 2022-08-04
Fix idx signup form btn-secondary color in footer.

B77Mills:
Screen Shot 2022-08-04 at 8 26 56 AM


jwade1327 (approved):

brandonbk (approved):

solocommand (commented):
Can't request changes now, but @B77Mills:

This doesn't work everywhere -- if you're intentionally adjusting the vertical spacing between the secondary button and the primary, you should ensure that the buttons are display block or flexed. This only works on PMMI because of the particular font sizes in place.
image

PR #413, opened 2022-08-24
Add ```marko-web-image-slider``` component

Shinsina:
After noticing this component has been essentially copy pasted (with essentially the only difference between any of them being the introduction of allowing an array of modifiers to be added), I opted to move it into Marko-Web so as to unify them as we continue to use this component go forward.

Should we rather include this in a different package or as a new one entirely that can be done, mostly just looking to alleviate having this copy pasted everywhere as it previously had been.

(Example of it generically being used as a Test on Minexpo)
Ascend-Test-Slider


Shinsina:
[Gallery Options](https://trello.com/c/1jfRbVXz/332-gallery-options)

Shinsina:
[Add image slider to all sites](https://trello.com/c/uRmnCqM9/338-add-image-slider-to-all-sites)

brandonbk (approved):

PR #423, opened 2022-09-09
Add support for skipping browser component rendering

zarathustra323:
When an existing instance of the component name is already present on the page.


B77Mills:
[Add Leaders package to global theme](https://trello.com/c/iWUivign/336-add-leaders-package-to-global-theme)

PR #433, opened 2022-09-23
Add support for disabling wallpaper, gtm, olytics

brandonbk:


solocommand:
@brandonbk Where/how is `disabledFeatures` defined and configured?

solocommand (approved):

PR #452, opened 2022-10-11
Add custom formatter support to Omeda+IdX rapid identification calls

B77Mills:
Add payload formatters to the onLoginLinkSent, onAuthenticationSuccess & onUserProfileUpdate.  This allows for individual groups to manipulate the payload being sent to omeda's rapidIdent process after the user hits submit & before it is sent to the rapid Ident call.  It will allow groups to adjust stuff like promoCodes and deploymentTypeIds being sent along.  

In this example we set a new appendPromoCode and auto opted the user into the deploymentTypeIds of 63.

Screen-Shot-2022-10-11-at-8 46 07-AM

### DIFF
--- a/packages/marko-web-omeda-identity-x/index.js
+++ b/packages/marko-web-omeda-identity-x/index.js
@@ -30,6 +30,9 @@ module.exports = (app, params = {}) => {
     omedaPromoCodeCookieName,
     omedaPromoCodeDefault,
     omedaRapidIdentifyProp,
+    onLoginLinkSentFormatter = (async ({ payload }) => ({ ...payload })),
+    onAuthenticationSuccessFormatter = (async ({ payload }) => ({ ...payload })),
+    onUserProfileUpdateFormatter = (async ({ payload }) => ({ ...payload })),
     rapidIdentProductId,
   } = validate(Joi.object({
     appendBehaviorToHook: Joi.array().items(schemas.hookBehavior),
@@ -93,6 +96,9 @@ module.exports = (app, params = {}) => {
     omedaPromoCodeCookieName: Joi.string().default('omeda_promo_code'),
     omedaPromoCodeDefault: Joi.string(),
     omedaRapidIdentifyProp: Joi.string().default('$omedaRapidIdentify'),
+    onLoginLinkSentFormatter: Joi.function(),
+    onAuthenticationSuccessFormatter: Joi.function(),
+    onUserProfileUpdateFormatter: Joi.function(),
     rapidIdentProductId: Joi.number().required(),
   }), params);
 
@@ -128,6 +134,9 @@ module.exports = (app, params = {}) => {
     omedaGraphQLClientProp,
     omedaPromoCodeCookieName,
     omedaPromoCodeDefault,
+    onLoginLinkSentFormatter,
+    onAuthenticationSuccessFormatter,
+    onUserProfileUpdateFormatter,
   });
 
   // attach the identity-x rapid identification wrapper middleware
--- a/packages/marko-web-omeda-identity-x/index.js
+++ b/packages/marko-web-omeda-identity-x/index.js
@@ -12,6 +12,8 @@ const rapidIdentifyRouter = require('./routes/rapid-identify');
 const props = require('./validation/props');
 const schemas = require('./validation/schemas');
 
+const defaultFormatter = async ({ payload }) => payload;
+
 module.exports = (app, params = {}) => {
   const {
     appendBehaviorToHook,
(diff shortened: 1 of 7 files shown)


packages/marko-web-omeda-identity-x/integration-hooks/on-login-link-sent.js
```diff
@@ -49,17 +51,20 @@ module.exports = async (params = {}) => {
...
       behavior,
       promoCode,
       appendBehaviors,
       appendDemographics,
       appendPromoCodes,
-    }),
-  ]);
+    },
+  });
+  const { encryptedCustomerId } = await idxOmedaRapidIdentify(payload);
```
solocommand: Give this a try instead to clean this up a bit:
```js
  const [omedaLinkedFields, { encryptedCustomerId }] = await Promise.all([
    getOmedaLinkedFields({ identityX, brandKey }),
    idxOmedaRapidIdentify(await formatter({
      req,
      payload: {
        user: user.verified ? user : { id: user.id, email: user.email },
        behavior,
        promoCode,
        appendBehaviors,
        appendDemographics,
        appendPromoCodes,
    })),
  ]);
```

packages/marko-web-omeda-identity-x/index.js
```diff
@@ -30,6 +30,9 @@ module.exports = (app, params = {}) => {
     omedaPromoCodeCookieName,
     omedaPromoCodeDefault,
     omedaRapidIdentifyProp,
+    onLoginLinkSentFormatter = (async ({ payload }) => ({ ...payload })),
+    onAuthenticationSuccessFormatter = (async ({ payload }) => ({ ...payload })),
+    onUserProfileUpdateFormatter = (async ({ payload }) => ({ ...payload })),
```
solocommand: Define your default formatter above (also no need to spread)
```js
const defaultFormatter = async ({ payload }) => payload;
```
And change this to use the function
```suggestion
    onLoginLinkSentFormatter = defaultFormatter,
    onAuthenticationSuccessFormatter = defaultFormatter,
    onUserProfileUpdateFormatter = defaultFormatter,
```

packages/marko-web-omeda-identity-x/index.js
```diff
@@ -93,6 +96,9 @@ module.exports = (app, params = {}) => {
     omedaPromoCodeCookieName: Joi.string().default('omeda_promo_code'),
     omedaPromoCodeDefault: Joi.string(),
     omedaRapidIdentifyProp: Joi.string().default('$omedaRapidIdentify'),
+    onLoginLinkSentFormatter: Joi.function(),
+    onAuthenticationSuccessFormatter: Joi.function(),
+    onUserProfileUpdateFormatter: Joi.function(),
```
solocommand, line 103: I believe these _should_ be required still. If someone explicitly passes `null` for `onLoginLinkSentFormatter`, this will fatal. Test this to make sure

solocommand (changes requested):

packages/marko-web-omeda-identity-x/index.js
B77Mills: @solocommand I don't believe it needs to be required here as it is setting the default value here and when it runs validation on the individual hooks using these it is doing a .function().required() check.  I believe at this step it need to set the default value or validate that what is being set is a function.  This would insure that at the second place this is used it wold have a function set at the key in the object.   If I am miss understanding let me know.  But as soon as i make it required at this point it doesn't function as I would expect for setting the default value.  Meaning if you set .required() it enforce that it be present so there is no point in setting the default value as you have to set it when you call the initial call.

services/example-website/config/omeda-identity-x.js
```diff
@@ -106,4 +108,63 @@ module.exports = {
...
+      // Get the encriptedCustomerId that matches the omeda brandKey
+      const encryptedCustomerId = user.externalIds.filter(({
+        identifier,
+        namespace,
+      }) => identifier.type === 'encrypted'
+      && namespace.provider === 'omeda'
+      && namespace.tenant === omeda.brandKey)[0].identifier.value;
+
+      // BAIL if no encryptedCustomerId and return payload
+      if (!encryptedCustomerId) return payload;
```
solocommand: Use an object-path helper here, if the user doesn't have any matching external ids, this will error. Also if you only want the first matching value, use `find` instead of `filter` because it will stop iterating once it finds a match.
```suggestion
      const found = getAsArray(user, 'externalIds')
        .find(({ identifier, namespace }) => identifier.type === 'encrypted'
          && namespace.provider === 'omeda'
          && namespace.tenant === omeda.brandKey
        );
      // BAIL if no encryptedCustomerId and return payload
      if (!found) return payload;
      const encryptedCustomerId = get(found, 'identifier.value');
```

solocommand (changes requested):

solocommand (approved):

B77Mills:
[Investigate when/how to add identityX hook to add newsletter signup based on users values within omeda](https://trello.com/c/r0j8lvjb/181-investigate-when-how-to-add-identityx-hook-to-add-newsletter-signup-based-on-users-values-within-omeda)

PR #455, opened 2022-10-13
Add product.type.id to getOmedaCustomerRecord for use with the new on${hook}formatters

B77Mills:
This will be used in conjuntion with the onLoginLinkSentFormatter to determine if the user has already been subscribed to the current website product.

The example-website example is using this to filter down the subscriptions based on current rapidIdentificationProductId set by the config.  If the user is already subscribed to the website product remove any promoCode being set for the `registration_meter`. This promoCode, or associated promoCode, should have already been set the first time they registered and subscribed to the website.


B77Mills:
[Investigate when/how to add identityX hook to add newsletter signup based on users values within omeda](https://trello.com/c/r0j8lvjb/181-investigate-when-how-to-add-identityx-hook-to-add-newsletter-signup-based-on-users-values-within-omeda)

solocommand (approved):

PR #461, opened 2022-10-24
Add configurable email validation to user creation

solocommand:
Allows a user to specify a custom email validation function. An example could be the following:
```js
const cfg = new IdentityXConfiguration({
  // ...
  emailValidator: async ({ email, req }) => {
    const { status, isValid } = await validate(email, req.ip);
    return [isValid, `Invalid: ${status}`];
  },
});
```
To explicitly abort the login process, throw an error in the email validator:
```js
const cfg = new IdentityXConfiguration({
  // ...
  emailValidator: async ({ email, req }) => {
    if (email === 'foo@bar.com') throw new Error('No foobars allowed.');
    return [true];
  },
});
```

### DIFF
--- a/packages/marko-web-identity-x/service.js
+++ b/packages/marko-web-identity-x/service.js
@@ -1,3 +1,4 @@
+const { get } = require('@parameter1/base-cms-object-path');
 const createClient = require('./utils/create-client');
 const getActiveContext = require('./api/queries/get-active-context');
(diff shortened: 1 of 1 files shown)


solocommand:
[Add ZeroBounce to IdentityX newsletter components](https://trello.com/c/hrAHdMT4/88-add-zerobounce-to-identityx-newsletter-components)

PR #467, opened 2022-10-31
Add new sideloadDataFn() to withContent & pageNode calls

B77Mills:
Add new sideloadDataFn to be passed into pageNode middleware.  This allows for additional data to be loaded along side the standard calls.

### DIFF
--- a/packages/marko-web/middleware/page-node/index.js
+++ b/packages/marko-web/middleware/page-node/index.js
@@ -9,12 +9,14 @@ class PageNode {
     queryFragment,
     variables,
     resultField,
+    sideloadDataFn,
   }) {
     this.apolloClient = apolloClient;
     this.queryFactory = queryFactory;
     this.queryFragment = queryFragment;
     this.variables = variables;
     this.resultField = resultField;
+    this.sideloadDataFn = sideloadDataFn;
   }
 
   async load() {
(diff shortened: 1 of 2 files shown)


packages/marko-web/middleware/page-node/index.js
```diff
@@ -9,21 +9,34 @@ class PageNode {
...
+          this.apolloClient.query({ query, variables }),
+          this.sideloadDataFn(),
+        ]).then(([r, sideloadedData]) => {
+          this.sideloadedData = sideloadedData;
+          return createNode(getAsObject(r, path));
+        });
+      } else {
+        this.promise = this.apolloClient.query({ query, variables })
+          .then(r => createNode(getAsObject(r, path)));
+      }
```
solocommand: Use `isFn` helper from base-cms-utils. Create a dummy function and always execute it, no need to duplicate the logic and nest if statements.

Don't bind your `this` context to the function -- you should not be exposing any internals of the page node middleware to the consuming function.

Don't mix async/await and then chains -- just use async/await.

```js
const fn = isFn(this.sideloadedDataFn) ? this.sideloadedDataFn : () => {};
const [r, data] = Promise.all([
  this.apolloClient.query(...),
  fn(),
];
this.promise = createNode(getAsObject(r, path));
this.sideloadedData = data;
```

solocommand (changes requested):

solocommand (approved):

PR #470, opened 2022-11-02
Include Sponsored content items in ```allCompanyContent``` resolver.

Shinsina:
Code examined to ensure intended behavior: https://github.com/parameter1/base-cms/blob/master/packages/utils/src/get-published-content-criteria.js -> https://github.com/parameter1/base-cms/blob/master/packages/utils/src/get-default-content-types.js -> https://github.com/parameter1/base-cms/blob/master/packages/utils/src/content-types.js and additionally https://github.com/parameter1/base-platform/blob/master/app/config/platform/models/modelspaces/platform/content/media/webinar.yml

Production:
Screen Shot 2022-11-02 at 11 48 07 AM

Dev:
Screen Shot 2022-11-02 at 11 48 25 AM

### DIFF
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -1247,7 +1247,7 @@ module.exports = {
         { 'relatedTo.$id': companyId },
       ];
 
-      const includeSponsored = (getAsArray(query, 'type.$in').includes('Event') || getAsArray(query, 'type.$in').includes('Webinar'));
+      const includeSponsored = getAsArray(query, 'type.$in').includes('Webinar');
       if (includeSponsored) query.$or.push({ sponsors: companyId });
 
       if (includeLabels.length && excludeLabels.length) {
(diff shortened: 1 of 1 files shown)


Shinsina:
[Displayed Sponsored Webinars on Company Pages](https://trello.com/c/LrOhDxt3/327-displayed-sponsored-webinars-on-company-pages)

solocommand (changes requested):
This change is fine, but it will require the creation of a new index on the Content collection to query for the `sponsors` field before it can be merged.

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -1247,6 +1247,9 @@ module.exports = {
         { 'relatedTo.$id': companyId },
       ];
 
+      const includeSponsored = getAsArray(query, 'type.$in').includes('Webinar');
+      if (includeSponsored) query.$or.push({ sponsors: companyId });
```
solocommand: Nevermind on the index, but lets just change this to include `sponsors` in the `$or` query above -- no need to add this separately.

solocommand (changes requested):

services/graphql-server/src/graphql/resolvers/platform/content.js
solocommand: Nevermind on the nevermind. Also need index.

solocommand (approved):

PR #471, opened 2022-11-02
Bail if crop dimensions aren’t available for "cropped" image

Shinsina:
Previously an assumption was being made (rightly or wrongly) that x1, x2, y1 and y2 would always be available on an Image that was marked as "cropped", however (as noted below) there are possibilities in the Base interface to crop an image without these values being set at all. Instead of returning null, causing the GraphQL query/fragment to fail for returning null on a non-nullable field, we will bail with the same logic we use when there is not a width, height or cropDimensions passed into this function.

URL: https://www.ccjdigital.com/alternative-power/article/15293351/truckings-ev-revolution-will-not-be-driven-by-price

Image from Production in current state:
Screen Shot 2022-11-02 at 2 14 27 PM

In Mongo:
Screen Shot 2022-11-02 at 2 32 48 PM

Asset causing issue in Mongo:
Screen Shot 2022-11-02 at 2 14 47 PM

Asset cause issue in Base:
Screen Shot 2022-11-02 at 2 23 46 PM

Resulting page following this correction:
Screen Shot 2022-11-02 at 2 34 26 PM

### DIFF
--- a/packages/image/src/crop-rectangle.js
+++ b/packages/image/src/crop-rectangle.js
@@ -55,6 +55,16 @@ module.exports = ({ width, height, cropDimensions }) => {
     return { ...o, [key]: v };
   }, {});
 
+  if (!x1 || !x2 || !y1 || !y2) {
+    return new CropRectangle({
+      x: 0,
+      y: 0,
+      width,
+      height,
+      cropped: false,
+    });
+  }
+
   return new CropRectangle({
     x: x1,
     y: y1,


Shinsina:
[CCJ: Getting a GraphQL error on story page](https://trello.com/c/JSCuQjsl/186-ccj-getting-a-graphql-error-on-story-page)

Shinsina:
[Photo Problem in Base](https://trello.com/c/cQ6lvUkr/330-photo-problem-in-base)

brandonbk (approved):

PR #485, opened 2022-11-09
Add transcript fields to podcast & videos both in the graphql & within monorail website

B77Mills:
To be done in conjunction with https://github.com/parameter1/base-platform/pull/183 && most likely a site specific site.  
 - Add transcript field to podcast and videos.
 - Create marko-web content element component for transcript field.
 - Add to Video structured data
 - Add to Podcast structured data
 
 
Screen Shot 2022-11-10 at 10 34 59 AM
Screen Shot 2022-11-10 at 10 35 09 AM

### DIFF
--- a/packages/marko-web/components/page/metadata/content.marko
+++ b/packages/marko-web/components/page/metadata/content.marko
@@ -33,10 +33,21 @@ fragment ContentPageMetadataFragment on Content {
   }
   ... on ContentVideo {
     embedSrc
+    metadata {
+      transcript
+    }
+  }
+  ... on ContentWebinar {
+    metadata {
+      transcript
+    }
   }
   ... on ContentPodcast {
     fileSrc
     fileName
+    metadata {
+      transcript
+    }
   }
   ... on Authorable {
     authors {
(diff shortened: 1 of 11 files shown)


solocommand:
[Support for video transcripts](https://trello.com/c/udmB4Lcn/162-support-for-video-transcripts)

packages/marko-web/components/page/metadata/content.marko
```diff
@@ -33,10 +33,12 @@ fragment ContentPageMetadataFragment on Content {
   }
   ... on ContentVideo {
     embedSrc
+    transcript
   }
   ... on ContentPodcast {
     fileSrc
     fileName
+    transcript
```
solocommand, line 41: This should also be added to `ContentWebinar` type

packages/marko-web/package.json
```diff
@@ -26,6 +26,7 @@
     "@parameter1/base-cms-tenant-context": "^3.0.0",
     "@parameter1/base-cms-utils": "^3.0.0",
     "@parameter1/base-cms-web-common": "^3.3.1",
+    "cheerio": "^0.22.0",
```
solocommand: Use one of the versions that are already installed in this project, `cheerio@^1.0.0-rc.10` or `cheerio@^1.0.0-rc.2`. This should have caused changes to the `yarn.lock` file, which I'm not seeing included here either.

packages/marko-web/components/page/metadata/google-structured-data/content.js
```diff
@@ -66,6 +73,7 @@ module.exports = (node) => {
       '@type': 'PodcastEpisode',
       headline: get(node, 'metadata.title'),
       ...(associatedMedia && { associatedMedia }),
+      ...(get(node, 'transcript') && { transcript: convertHtmlToText(get(node, 'transcript')) }),
```
solocommand: Missing field for `Webinar`s

packages/marko-web/components/page/metadata/google-structured-data/content.js
```diff
@@ -49,6 +55,7 @@ module.exports = (node) => {
       uploadDate: publishedISOString,
       contentUrl: get(node, 'siteContext.canonicalUrl'),
       embedUrl: get(node, 'embedSrc'),
+      ...(get(node, 'transcript') && { transcript: convertHtmlToText(get(node, 'transcript')) }),
```
solocommand: What happens if the user puts in a bad or empty value? If this throws, does the entire page error or does the JSON+LD data break quietly?

solocommand (changes requested):

packages/marko-web/components/page/metadata/google-structured-data/content.js
B77Mills: @solocommand  I added a value check to the converthHtmlToSring function & I also added a check to make sure there is a returned value.  In testing or trying to make it break I was only able to make cheerio break by not sending a value at all.  It converted broken or crappy html just fine.  I can add some additional check but not sure if you want me to wrap the cheerio.load() in some sort of try catch?  Thoughts?

packages/marko-web/components/page/metadata/google-structured-data/content.js
B77Mills: @solocommand This was intentional until we have clearer direction into what type of object a webinar needs to be and what fields need to be add along with the new transcript field.

solocommand (approved):

PR #487, opened 2022-11-10
Adjust embedded image targeting for bottom border

Shinsina:
Production:
Screenshot from 2022-11-10 13-36-51

Dev:
Screenshot from 2022-11-10 13-31-56

NOTE: I opted not to change the behavior of https://github.com/parameter1/base-cms/blob/master/packages/marko-web-theme-monorail/scss/components/_content-page.scss#L263 as that behavior is functionally/logically different than these, if that should also be changed to be targeting in the same fashion, leave a review noting it.

### DIFF
--- a/packages/marko-web-theme-default/scss/components/_page-image.scss
+++ b/packages/marko-web-theme-default/scss/components/_page-image.scss
@@ -92,7 +92,9 @@
 }
 
 [data-embed-type="image"] {
-  > :last-child:not(img) {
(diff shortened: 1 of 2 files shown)


Shinsina:
[When images are uploaded in Base 'extra' code is also being added.](https://trello.com/c/5WVgRc0L/392-when-images-are-uploaded-in-base-extra-code-is-also-being-added)

brandonbk (approved):

PR #488, opened 2022-11-11
Add profile onSubmit message and redirect logic

B77Mills:
On submit hide the form and display a success message.  The message will be one of the two following message depending on if there is a returnTo url set or not.

### With returnTo:
Your profile has been saved.
If you are not automatically redirected click here to continue.
Screen Shot 2022-11-11 at 10 15 00 AM

### Without returnTo:
Your profile has been saved.
Click here to continue modiying your profile or click here to return to the home page.
Screen Shot 2022-11-11 at 10 14 38 AM

### DIFF
--- a/packages/marko-web-identity-x/browser/profile.vue
+++ b/packages/marko-web-identity-x/browser/profile.vue
@@ -185,9 +185,8 @@
         <div v-else class="success-message__message">
           <span class="btn-link" @click="handleReload()">
             Click here
-          </span> to continue modiying your profile or <a href="/">
-            click here
-          </a> to return to the home page.
+          </span> to continue modiying your profile or
+          <a href="/">click here</a> to return to the home page.
         </div>
       </div>
     </div>
--- a/packages/marko-web-identity-x/browser/profile.vue
+++ b/packages/marko-web-identity-x/browser/profile.vue
@@ -1,7 +1,7 @@
 <template>
   <div v-if="hasActiveUser">
-    <p>{{ callToAction }}</p>
-    <form @submit.prevent="handleSubmit">
+    <p v-if="!didSubmit">{{ callToAction }}</p>
+    <form v-if="!didSubmit" @submit.prevent="handleSubmit">
       <fieldset :disabled="isLoading">
         <div class="row">
           <div
@@ -176,6 +176,20 @@
         An error occurred: {{ error }}
       </p>
     </form>
+    <div v-else>
+      <div class="success-message">
+        <div class="success-message__title">
+          Your profile has been saved.
+        </div>
+        <div v-if="returnTo" class="success-message__message">
+          If you are not automatically redirected <a :href="returnTo">click here</a> to continue.
+        </div>
+        <div v-else class="success-message__message">
+          <a href="#" @click="handleReload()">Click here</a> to continue modiying your profile
+          or <a href="/">click here</a> to return to the home page.
+        </div>
+      </div>
+    </div>
   </div>
   <div v-else>
     <p>You must be logged-in to modify your user profile.</p>
@@ -321,6 +335,10 @@ export default {
     returnTo: {
       type: String,
       default: null,
+    },
+    returnToDelay: {
+      type: Number,
+      default: 5000,
     },
     enableChangeEmail: {
       type: Boolean,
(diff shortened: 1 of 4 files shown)


packages/marko-web-theme-monorail/scss/components/_identity-x.scss
```diff
@@ -34,4 +34,15 @@
       border-radius: 4px;
     }
   }
+  &--profile {
+    .success-message {
+      color: pink;
```
solocommand: Why pink? just use `text-success` class and allow the theme to dictate what that color actually is.

packages/marko-web-identity-x/browser/profile.vue
```diff
@@ -176,6 +176,20 @@
...
+      <div class="success-message">
+        <div class="success-message__title">
+          Your profile has been saved.
+        </div>
+        <div v-if="returnTo" class="success-message__message">
+          If you are not automatically redirected <a :href="returnTo">click here</a> to continue.
+        </div>
+        <div v-else class="success-message__message">
+          <a href="#" @click="handleReload()">Click here</a> to continue modiying your profile
+          or <a href="/">click here</a> to return to the home page.
```
solocommand: Make this a button with `btn-link` class rather than an A tag with a search param. `modiying` should be `modifying`, and I'd recommend this text:
> [Click here]() to make further changes to your preferences, or [return to the home page]().

packages/marko-web-identity-x/browser/profile.vue
```diff
@@ -322,6 +336,10 @@ export default {
       type: String,
       default: null,
     },
+    returnToDelay: {
```
solocommand, line 339: This must be defined and passed through in both the `marko` components, and any other components that include the profile (such as Authenticate, Login, and Comment)

packages/marko-web-identity-x/browser/profile.vue
```diff
@@ -176,6 +176,20 @@
         An error occurred: {{ error }}
       </p>
     </form>
+    <div v-else>
+      <div class="success-message">
+        <div class="success-message__title">
+          Your profile has been saved.
+        </div>
+        <div v-if="returnTo" class="success-message__message">
+          If you are not automatically redirected <a :href="returnTo">click here</a> to continue.
```
solocommand: > You will be automatically redirected in {{ returnToDelay / 1000 }} seconds. [Click here]() to continue.

solocommand (changes requested):
From the UX side, 5 seconds is a **long** time to wait after the form has been submitted. While I don't like the idea of the form hiding in the first place, making the user wait 5 seconds to continue (especially when this is only an intermediate step such as inline gated content or redirect to external for LME) seems frustrating.

Regarding the form collapse: This will incur a lot of motion, and may end up being _more_ frustrating for the user. The 'submitted' message will flash for a frame or so next to the submit button, then immediately disappear and scroll to the top of the page (due to the length of the form no longer pushing them down). This in turn will cause the footer to take up most of the screen.

The success message should be appended to the bottom of the form, and should not cause elements to disappear and jump around on the page.

packages/marko-web-identity-x/browser/profile.vue
B77Mills: @solocommand  I did add this to marko.json and added passing it through the marko component if it is sent along with the input.   

As for adding it to the other things mentioned. The commenting & login component do not call or reference the profile component and the auth component is overriding the submit(profile-update) action & forcing a redirect already [line 27](https://github.com/parameter1/base-cms/blob/master/packages/marko-web-identity-x/browser/authenticate.vue)

I don't think this is currently being called in any other spots.

B77Mills: @solocommand  I also checked the gated content part which should be using the reload-page-on-submit=true prop which would force the content page to reload on profile submit.  But in theory the user should be intercepted at the authentication step to finish filling out the required fields and then get redirected to the content landing page.  If they did some how get to the content page without a fully filled out profile page this would still reload the page on submission of a full profile.

solocommand (approved):

B77Mills:
[On profile update, redirect to homepage if no redirect URL is present](https://trello.com/c/4Y2LZzWZ/105-on-profile-update-redirect-to-homepage-if-no-redirect-url-is-present)

PR #491, opened 2022-11-16
Add emptyHTML check to only return transcripts that are not empty 

B77Mills:
This will insure that the transcript field is actually set to something with text in it vs empty html.  Before if you cleared out or the user has just save a couple of empty html element it would return and display what looked to be nothing, but was really emply p  & h tags.  This check looks to see if there is actually text to render within the given html elements.

For example if you clear out the transcript field or hit return once in the field and save the model you now have an empty <p></p> saved in the value for transcript.  This new check would see that value as empty when it convert the html to text and trim the whitespace.  It will return a null value instead of an empty html string.

This should allow for this [PR](https://github.com/parameter1/industrial-media-websites/pull/287) to be closed, but the site repos will have to have a dep upgrade done to prevent empty transcript fields from displaying.

### DIFF
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -63,7 +63,7 @@ const retrieveYoutubePlaylistId = async ({ youtube }) => {
   }
 };
 
-const convertEmptyHtmlToNull = async (html) => {
+const convertEmptyHtmlToNull = (html) => {
   if (!html) return null;
   const $ = cheerio.load(html);
   if (!$.text().trim()) return null;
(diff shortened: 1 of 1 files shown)


B77Mills:
[Add new transcript field to remaining sites.](https://trello.com/c/plo9g5C6/150-add-new-transcript-field-to-remaining-sites)

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -809,6 +816,30 @@ module.exports = {
...
+    transcript: async (content) => {
+      const transcript = await convertEmptyHtmlToNull(content.transcript);
+      return transcript;
+    },
+  },
+
+  /**
+   *
+   */
+  ContentPodcast: {
```
zarathustra323, line 825:
```js
/**
 *
 */
ContentPodcast: {
  transcript: ({ transcript }) => convertEmptyHtmlToNull(transcript),
},
```

services/graphql-server/src/graphql/resolvers/platform/content.js
```diff
@@ -63,6 +63,13 @@ const retrieveYoutubePlaylistId = async ({ youtube }) => {
   }
 };
 
+const convertEmptyHtmlToNull = async (html) => {
```
zarathustra323: Remove `async`

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #493, opened 2022-11-21
Verify GAM exists before calling methods on it

Shinsina:
This is contained within a feed block of the Monorail theme. In the case provided this is https://github.com/parameter1/base-cms/blob/master/packages/marko-web-theme-monorail/components/blocks/latest-products-feed.marko which is a part of products layout for AB, provided that AB is no longer using GAM calling GAM "willy nilly" like this causing erroring as GAM is no longer defined in that context. 

Provided that the block calling this is a member of the theme (see previous link), this should likely be accounted for anywhere that is utilizing this component to allow for it's "GAM logic" to be replaced by passed in components from the site level if need be.

One could argue that these components shouldn't be a part of `theme-monorail` (or at the very least not part of it's core implementation) and should be ported into either in place of the existing versions in `marko-web-gam` (or in addition to the ones already in `marko-web-gam` or into it's own package of `marko-web-theme-monorail-gam` etc. as while it may be typical that a site will use GAM in the even it doesn't it shouldn't inhibit the use of the items contained within the theme itself.

Production:
Screenshot from 2022-11-21 11-33-00

Development:
Screenshot from 2022-11-21 11-30-13

### DIFF
--- a/packages/marko-web-theme-monorail/components/gam/define-display-ad.marko
+++ b/packages/marko-web-theme-monorail/components/gam/define-display-ad.marko
@@ -6,7 +6,7 @@ import defaultValue from "@parameter1/base-cms-marko-core/utils/default-value";
 
 $ const { GAM, req } = out.global;
 
-$ const adUnit = GAM.getAdUnit({ name: input.name, aliases: input.aliases });
+$ const adUnit = GAM ? GAM.getAdUnit({ name: input.name, aliases: input.aliases }) : {};
 $ const { templateName } = adUnit;
 $ const showLabel = defaultValue(input.showLabel, true);
 $ const targeting = defaultValue(input.targeting, {});


Shinsina:
[AQ Error on Website](https://trello.com/c/dhEbuX2w/395-aq-error-on-website)

packages/marko-web-theme-monorail/components/gam/define-display-ad.marko
```diff
@@ -6,7 +6,7 @@ import defaultValue from "@parameter1/base-cms-marko-core/utils/default-value";
 
 $ const { GAM, req } = out.global;
 
-$ const adUnit = GAM.getAdUnit({ name: input.name, aliases: input.aliases });
+$ const adUnit = GAM ? GAM.getAdUnit({ name: input.name, aliases: input.aliases }) : {};
 $ const { templateName } = adUnit;
```
brandonbk, line 10: @Shinsina - if GAM doesn't exist, and templateName doesn't exist, what happens?

brandonbk (changes requested):

packages/marko-web-theme-monorail/components/gam/define-display-ad.marko
Shinsina: @brandonbk See production screenshot, it hard fails because if GAM doesn't exist GAM.getAdUnit isn't a valid call as "cannot read property getAdUnit of undefined"


EDIT: If you're asking about behavior after the change I believe it will throw a console.warn that it can't render an ad unit as no path was provided (as adUnit is an empty object)

solocommand (changes requested):
This should instead be changed to check if GAM is enabled _before_ calling this component. In the case of the theme block (latest products), the gam define display block should only be called if GAM is enabled. The block could have an input to determine if ad serving should be disabled. If that doesn't make sense for that block, the implementing site should likely override this component to customize ad delivery.

The warning about the ad unit missing a path is supposed to inform the developer of a misconfiguration, not something to be ignored.

Shinsina:
@solocommand Closing in favor of the already active solution: https://github.com/parameter1/ab-media-websites/pull/246 upon additional review this would need to get changed in multiple page templates within `marko-web-theme-monorail` and `marko-web-theme-magazine` that are likely to be overwritten in the event GAM isn't available. 

That does present the potential of inconsistent presentation/behavior of these pages/elements when overwritten at the site/tenant level, additionally falling back into the same concerns I had regarding this behavior being present to begin with.

This makes the assumption that GAM exists and is present and while yes `marko-web-gam` is listed as a dependency, my thought would be it shouldn't be required _as part of the core implementation_ there isn't anything that couldn't be reworked to include a slot to slide the ads in wherever they're supposed to appear (Such as here: https://github.com/parameter1/base-cms/blob/master/packages/marko-web-theme-monorail/templates/search.marko#L36) but seeing as there isn't any other sites utilizing this package currently (aside from the AB ones that already account for it via overrides) that utilize a different ad service than GAM, I've opted to continue the same way it's been accounted for via component/template overriding as you had also suggested if it made more sense to do it that way.

PR #459, opened 2022-10-17
Support additional search page translations

Shinsina:
Search-Page-1
Search-2
Search-Lower
Filters-Search

Screenshots from BLP (A site without an i18n function definition):
Screen Shot 2022-11-03 at 9 26 03 AM
Screen Shot 2022-11-03 at 9 26 09 AM
Screen Shot 2022-11-03 at 9 26 34 AM
Screen Shot 2022-11-03 at 9 26 40 AM
Screen Shot 2022-11-03 at 9 26 51 AM
Screen Shot 2022-11-03 at 9 27 17 AM

Screenshots from AW (A site with an i18n function definition but is not translated (remains the same as before)):
Screen Shot 2022-11-03 at 9 30 16 AM
Screen Shot 2022-11-03 at 9 30 20 AM
Screen Shot 2022-11-03 at 9 30 40 AM
Screen Shot 2022-11-03 at 9 30 45 AM
Screen Shot 2022-11-03 at 9 30 54 AM
Screen Shot 2022-11-03 at 9 30 59 AM

PR #497, opened 2022-11-30
Replace moment with dayjs, support `es` and `en` locales and translate 

Shinsina:
Utilized the same check for if i18n is configured as https://github.com/parameter1/base-cms/pull/459 which should also work here as the Allured search pages still work (which were the test cases in that PR for such a condition).

Screenshot from 2022-11-30 10-54-27
Screenshot from 2022-11-30 10-56-00
Screenshot from 2022-11-30 10-56-04


Shinsina:
[Mundo site translations needed](https://trello.com/c/d5s711ly/478-mundo-site-translations-needed)

Shinsina:
Closing in favor of https://github.com/parameter1/base-cms/pull/498 which includes the changes that would have been made here.

PR #506, opened 2022-12-05
Add ability to pass in component modifiers & selector

B77Mills:
Add ability to passing modifiers to the component being rendered.

Add ability to pass in a selector and use it as the component ID and set as the attrs.id prop on the included component.


B77Mills:
[Convert remaining monorail theme content body's to use new ad injection component](https://trello.com/c/sL1Oc2TV/12-convert-remaining-monorail-theme-content-bodys-to-use-new-ad-injection-component)

brandonbk (approved):

PR #511, opened 2022-12-07
Update htmlBodyInjector selector to exclude p with last element being a `<span data-embed-type />`

B77Mills:
Update logic to skip any p tag that has an span for embed as its last element.  This doesn mean that it will no longer count the copy in those p tags towards the word count.

**Here are a couple of DEV(left) vs PROD(right)  screen shots:**
Screen Shot 2022-12-07 at 10 19 32 AM

Screen Shot 2022-12-07 at 10 18 52 AM


B77Mills:
[Weird text wrapping around an image](https://trello.com/c/eUoZUxuz/500-weird-text-wrapping-around-an-image)

PR #528, opened 2022-12-28
Add new source cookie util & add set call to IdentityX login route

B77Mills:
This will use the login source being sent to the login component to set the cookie value for login source.  This will be used at later steps in the identity x process to send along source information.   

In this example it will now set the identity-x-source: comments so that later on when the user authenticates or updates their profile it would know that the login source was commenting and can adjust the promoCode &or products accordingly.


B77Mills:
[Implement "article_comments" promo code](https://trello.com/c/ENuNJ42c/153-implement-articlecomments-promo-code)

PR #537, opened 2023-01-19
IdentityX user fragments, custom attribute support

solocommand:
Adds a method to the IdentityX service to mutate the user's custom attribute data.

Adds preview mode to example site user profile page to print custom data and increment a counter.


solocommand:
[Send & retrieve `customAttributes`](https://trello.com/c/Z6BmGvXU/27-send-retrieve-customattributes)

zarathustra323 (commented):
I need some time to digest this. Perhaps a call is in order? At first blush I don't think this needs to be as "complicated" as it is. Why not handle custom attributes all the time?

zarathustra323 (approved):

PR #552, opened 2023-02-06
Allow overriding of link destination in `theme-section-feed-content-node`

Shinsina:
Showing an override on the Webinar type with the `linkUrl` field
Screenshot from 2023-02-06 13-10-58

Showing that default behavior is preserved on non-overridden types:
Screenshot from 2023-02-06 13-11-12


Shinsina:
[Webinar: Ability to Link to external form from landing page.](https://trello.com/c/aUP0jUiI/531-webinar-ability-to-link-to-external-form-from-landing-page)

B77Mills (commented):
@Shinsina & @brandonbk  The only thing I see coming up with this is off site targeting not opening in a blank window.  I would prefer that it doesn't, but guessing the group will ask about it.  

I believe that is part of the linkAttrs, but not sure how to add this via config or programmatically.
 
Either of you have a preference or opinion on how that should work?

packages/marko-web-theme-monorail/components/nodes/section-feed-content.marko
```diff
@@ -17,6 +17,16 @@ $ const withSection = defaultValue(input.withSection, true);
 $ const withTeaser = defaultValue(input.withTeaser, true);
 $ const displayImage = defaultValue(input.displayImage, true);
 $ const lazyload = defaultValue(input.lazyload, true);
+// linkOffsite is an object like the following:
```
B77Mills, line 20: @brandonbk  & @Shinsina opinion question on this one.  This will work or an alternative is allow for a link prop to be passed in which would set these from whichever component is call it 

This file would need this added instead
```markojs
<!-- set link values and attrs based on input -->
$ const link = {
  ...(linkAttrs &&  { attrs: linkAttrs }),
  ...(input.link && { ...getAsObject(input, "link") })
};
<!-- set href value for image based of above ink props -->
$ const href = (link && link.field) ? get(content, link.field) : get(content, "siteContext.path");
```

link={ field: linkField, attrs: linkAttrs } would become link=link
& 
<@link href=link attrs=linkAttrs /> would become something link <@link href=href attrs=linkAttrs />

The part this still doesn't address is the targeting of the link base on external vs internal, togging base on if linkUrl is set.  

Example call:
```markojs
<theme-section-feed-block alias=alias modifiers=["upcoming"]>
  <@header>
    Upcoming
  </@header>
  <@query-params ...upcomingQueryParams />
  <@node link={ field: 'linkUrl' } linkAttrs={ target: '_blank' } />
  <@after>
    <theme-gam-define-display-ad
      name="rotation"
      position="section_page"
      aliases=aliases
      modifiers=["inter-block"]
    />
  </@after>
</theme-section-feed-block>
```

Shinsina: @B77Mills https://github.com/parameter1/base-cms/pull/552/commits/45dc2af16fde7672a6530be48f4e779a09094a5a The changes here should allow for the toggling of `target="_blank"` per site, think it's important to assure this behavior is consistent for anywhere this component is utilized as that's ultimately the "goal" is to drive traffic to these links when configured so regardless of the site location it should use this behavior

B77Mills (approved):

PR #569, opened 2023-02-13
Add additionalEventData flag for createdNewUser

B77Mills:
Add new additionEventData.createdNewUser  that is set when IdentityX creates a new user along with passing appUser to the loginLinkSent response.  This will allow for these to be passed through to the frontend to send event to gtm.


B77Mills:
[Pass newsletter signup events to GA4](https://trello.com/c/FTkD60CR/210-pass-newsletter-signup-events-to-ga4)

brandonbk (approved):

PR #574, opened 2023-02-16
Query for redirect `from` values containing the input `params`

Shinsina:
Previously this wasn't trying to look for `from` values with the passed in query string parameters from the query input, this does that and redirects the item with the existing behavior.

Example: http://www-policemag.dev.parameter1.com:9801/tags?tag=wisconsin would now go to
http://www-policemag.dev.parameter1.com:9801/t/6817018?tag=wisconsin

Screenshot from 2023-02-16 09-56-37

This will not require any changes to dependencies or site routing.

### DIFF
--- a/packages/marko-web-identity-x/browser/login.vue
+++ b/packages/marko-web-identity-x/browser/login.vue
@@ -16,18 +16,7 @@
       {{ buttonLabels.logout || "Log out" }}
     </a>
   </div>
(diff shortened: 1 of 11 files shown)


Shinsina:
[Tag redirects](https://trello.com/c/Hmi4EaQY/75-tag-redirects)

PR #617, opened 2023-03-01
Assert that `cropRectangle` returns `coords >= 0`

Shinsina:
Ref: https://docs.imgix.com/apis/rendering/size/rect 

Imgix doesn't support crop dimensions less than 0 so we must assert that these are at least 0.


Shinsina:
[Photo problem in newsletters](https://trello.com/c/R7a6yH7E/363-photo-problem-in-newsletters)

zarathustra323 (approved):

PR #621, opened 2023-03-03
"Whitelist" video embed providers and assert them as `data-embed-type="video"`

Shinsina:
Previously many video embeds were coming back as links, this attempts to resolve a decent number of the ones "missed" and asserts them based on the provider name AND specific keywords (player or video) in the URL of the embed. This was done so as to not change all embeds from certain providers to videos etc.

Live:
Screenshot from 2023-03-02 21-54-00

Dev:
Screenshot from 2023-03-02 21-54-03
Screenshot from 2023-03-02 21-54-18

### DIFF
--- a/services/oembed/src/app.js
+++ b/services/oembed/src/app.js
@@ -93,12 +93,12 @@ const videoProvider = new Set([
   'wwltv.com',
 ]);
 
-const retrieveOembed = async ({url, params}) => {
+const retrieveOembed = async ({ url, params }) => {
   const data = await embedly.oembed(url, params);
   // Match on "valid video urls" and enforce providers.
   if (url.match(/player|video/) && videoProvider.has(data.provider_name)) return { ...data, type: 'video' };
   return data;
-}
+};
 
 app.post('/', asyncRoute(async (req, res) => {
   const { url, ...params } = req.body;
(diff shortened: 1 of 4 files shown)


Shinsina:
[Video embed getting cut off](https://trello.com/c/EWeQG3Fc/109-video-embed-getting-cut-off)

zarathustra323 (changes requested):
The issue is that, somehow, the video was added with a wrapping `data-oembed-type="link"` instead of `data-oembed-type="video"`

If changed to the latter the aspect ratio is applied.

Current (note the oembed type of `link`):
image

After changing the `data-oembed-type` to `video`:
image

services/oembed/src/app.js
```diff
@@ -21,9 +21,88 @@ app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal']);
...
+  'Viddler',
+  'wcnc.com',
+  'wthr.com',
+  'wwltv.com',
+]);
+
+const retrieveOembed = async ({ url, params }) => {
+  const data = await embedly.oembed(url, params);
+  // Match on "valid video urls" and enforce providers.
+  if (url.match(/player|video/) && videoProvider.has(data.provider_name)) return { ...data, type: 'video' };
```
zarathustra323: Should this be an `&&` or an `||`?

zarathustra323 (changes requested):

zarathustra323 (approved):

PR #663, opened 2023-03-31
Allow Leaders Contextual to load sections by content schedules.

Shinsina:
The comment `// Look up leadership sections by content taxonomy, scheduling, and primary section` is presently false as the respective function `loadContentSections` was only accounting for taxonomy and primary section as shown due to the necessary changes to the fragment found in https://github.com/parameter1/base-cms/blob/master/packages/leaders-program/src/graphql/queries/content.js. This now corrects this and allows for the functionality to be opted into using the `useContentSchedules` prop with defaults to false (exactly like `useContentPrimarySection`)

PRODUCTION:
Screenshot from 2023-03-31 09-34-11

DEVELOPMENT:
Screenshot from 2023-03-31 09-34-14

### DIFF
--- a/packages/leaders-program/src/components/leaders.vue
+++ b/packages/leaders-program/src/components/leaders.vue
@@ -328,7 +328,7 @@ export default {
         const scheduledSections = getAsArray(r1, 'data.content.websiteSchedules');
         if (scheduledSections.length) {
           sectionIds.push(
-            ...scheduledSections.map((scheduledSection) => scheduledSection.section.id)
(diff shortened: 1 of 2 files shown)


Shinsina:
[Allow Leaders mapping to read from scheduling vs primary site/section](https://trello.com/c/tbE2tRyS/472-allow-leaders-mapping-to-read-from-scheduling-vs-primary-site-section)

B77Mills (approved):

PR #664, opened 2023-04-04
Add gated content support for  google structured data

B77Mills:
Add the ability to get the content gating function from out.global or use the defaultFn which just looks at the nodes 'userRegistration.isCurrentlyRequired' property.

By default nothing should change.  In the case of AB it will utilize the global function to apply the content gating of all article content types.

Screen Shot 2023-04-04 at 10 36 13 AM

Screen Shot 2023-04-04 at 10 36 39 AM

### DIFF
--- a/packages/marko-web/components/page/metadata/content.marko
+++ b/packages/marko-web/components/page/metadata/content.marko
@@ -31,6 +31,9 @@ fragment ContentPageMetadataFragment on Content {
       src(input: { options: { auto: "format,compress", w: "1200", fit: "max", q: 70 } })
     }
   }
+  userRegistration {
+    isCurrentlyRequired
+  }
   ... on ContentVideo {
     embedSrc
     transcript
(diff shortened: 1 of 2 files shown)


brandonbk (dismissed):

packages/marko-web/components/page/metadata/content.marko
```diff
@@ -87,6 +90,9 @@ ${processedFragment}
 `;
 
 $ const buildStructuredData = isFunction(input.buildStructuredData) ? input.buildStructuredData : defaultBuildStructuredData;
+$ const defaultFn = ({ content }) => get(content, 'userRegistration.isCurrentlyRequired', false);
```
solocommand: This should not destructure `content` from the first argument, or L50 needs to change to send `{ content: node }` instead of `node` as the first argument. Not sure which one is correct -- where is the contentGatingHandler definition?

solocommand (changes requested):

solocommand (approved):

PR #673, opened 2023-04-07
Move change email button and display address

solocommand:
If enabled, display the user's current email address and move the "change" button next to it.

image
image


solocommand:
[Can we display the users email field on the user profile page when someone signs in?](https://trello.com/c/mgbN1dLG/241-can-we-display-the-users-email-field-on-the-user-profile-page-when-someone-signs-in)

zarathustra323 (approved):

PR #676, opened 2023-04-11
Add regByPass & contentGatingHandler support to content metering middleware

B77Mills:
Account for when content is already gated by reg, either by default or by the global contentGatingHandler.  In this case if it is already gated return showGate of false as it will already force the user to fill out registration to access.


B77Mills:
[Gate all 'Article' content types for Athletic Business only](https://trello.com/c/17UyjmlC/446-gate-all-article-content-types-for-athletic-business-only)

solocommand (approved):

PR #678, opened 2023-04-13
Use `autocomplete.phrase` when configured appropriately.

Shinsina:
This is not an end all be all "magic silver bullet" regarding search however this at the very least better lines up the entered input with the returned results.

PRODUCTION:
Screenshot from 2023-04-13 13-17-48

DEVELOPMENT:
Screenshot from 2023-04-13 13-17-00

We may however want to investigate how "boosts", sorts, etc. are applied in either case (see: https://github.com/parameter1/base-browse/blob/main/api/src/graphql/search-content.js#L234)

### DIFF
--- a/packages/marko-web-search/components/query.marko
+++ b/packages/marko-web-search/components/query.marko
@@ -22,7 +22,7 @@ $ const params = {
   ),
   queryFragment: input.queryFragment,
   opSuffix: input.opSuffix,
-  enableNewSearch: site.get('enableNewSearch', false),
+  useAutocomplete: site.get('useAutocomplete', false),
 };
 
 <marko-web-resolve|{ resolved }| promise=loader({ apolloBaseCMS, apolloBaseBrowse }, params)>
--- a/packages/marko-web-search/components/query.marko
+++ b/packages/marko-web-search/components/query.marko
@@ -4,6 +4,7 @@ $ const {
   apollo: apolloBaseCMS,
   $baseBrowse: apolloBaseBrowse,
   config,
+  site
 } = out.global;
 $ const { assignedToWebsiteSectionIds } = input;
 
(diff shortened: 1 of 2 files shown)


Shinsina:
[Search: Enable Exact Phrase Search](https://trello.com/c/3VXxkC6f/451-search-enable-exact-phrase-search)

packages/marko-web-search/loaders/search.js
```diff
@@ -66,12 +66,15 @@ module.exports = async ({ apolloBaseCMS, apolloBaseBrowse } = {}, {
   opSuffix,
 } = {}) => {
   if (!apolloBaseCMS || !apolloBaseBrowse) throw new Error('Both the BaseCMS and Base Browse Apollo clients must be provided.');
+  const query = (!searchQuery.match(/^".+"$/) && !searchQuery.match(/^'.+'$/)) ? searchQuery : '';
+  const phrase = (searchQuery.match(/^".+"$/) || searchQuery.match(/^'.+'$/)) ? searchQuery : '';
```
B77Mills: @Shinsina I believe you will also want to strip the wrapping quotes before setting searchQuery here. 
something like 
```markoJs
const phrase = (searchQuery.match(/^".+"$/) || searchQuery.match(/^'.+'$/)) ? searchQuery.replace(/^["'](.+(?=["']$))["']$/, '$1') : '';
```

solocommand (changes requested):
@Shinsina Change this to use `useAutocomplete` as the parameter instead of `enableNewSearch`, as that's more descriptive of what is actually happening here.

packages/marko-web-search/components/query.marko
```diff
@@ -22,7 +22,7 @@ $ const params = {
   ),
   queryFragment: input.queryFragment,
   opSuffix: input.opSuffix,
-  enableNewSearch: site.get('enableNewSearch', false),
+  useAutocomplete: site.get('useAutocomplete', false),
```
solocommand: One more change, key this under `search`:
```suggestion
  useAutocomplete: site.get('search.useAutocomplete', false),
```

solocommand (changes requested):

PR #679, opened 2023-04-17
Adjust Identity X events emitter

B77Mills:
Adjust how Identity X event emitter handles loginSource, source & actionSource.  
```js
const source = this.loginSource || this.source;
const dataActionSource = data ? data.actionSource : undefined;
const actionSource = dataActionSource || window.IdentityX.getLoginSource() || source;
```
It will then also set actionSource, loginSource & source to the calculated actionSource. 

I also add new mixin to the identityX package to handle the event for the auto signup from additionalEvent Data.  This specifically deals with the identity-x-auto-signup event.

### DIFF
--- a/packages/marko-web-identity-x/browser/mixins/global-auto-signup-event-emitter.js
+++ b/packages/marko-web-identity-x/browser/mixins/global-auto-signup-event-emitter.js
@@ -1,4 +1,4 @@
-import getAsArray from '@parameter1/base-cms-object-path';
+import getAsArray from '../../utils/get-as-array';
 
 export default {
   inject: ['EventBus'],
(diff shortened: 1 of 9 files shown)


solocommand (changes requested):
Don't create a getAsArray util here, and no need to import `object-path`. We're already importing `@parameter1/base-cms-object-path`, just use that.

The IdentityX event registrations should move from the `marko-web-gtm` package into the `marko-web-identity-x` package, and the override of `loginSource` to the dataLayer should be removed.

Do we need to distinguish between an auto-signup emitted from login vs authenticate? If so, we should either pass the event name as part of the payload, or send distinct events (`identity-x-auto-signup-from-login` vs `identity-x-auto-signup-from-authenticate`).

Looks good otherwise!

packages/marko-web-identity-x/browser/mixins/global-auto-signup-event-emitter.js
```diff
@@ -0,0 +1,18 @@
+import getAsArray from '@parameter1/base-cms-object-path';
```
solocommand: Pretty sure this should be `import { getAsArray } ...` :)

solocommand (changes requested):

solocommand (approved):

PR #732, opened 2023-06-26
Update how default browser config values are assigned

B77Mills:
This update will ensure that all browser config values have a their default values applied.

I also removed the withGTM prop as it was no longer referenced and config was also not passed through anywhere.


solocommand (changes requested):
@B77Mills I think this is fine, assuming that we _want_ the defaults to be applied to everyone. That said, `idxArgs` and `inquiryArgs` still need to have a default/fallback to an empty object (in case someone sends `Monorail(Browser, { idxArgs: null })`).

Update lines 89, 91, and 94 to use `idxArgs || {}` and `inquiryArgs || {}`.

B77Mills:
@solocommand  I updated the setting of those values to use getAs object to ensure they are, at minimum and empty, object.

solocommand (approved):

PR #733, opened 2023-06-28
Add verifiedCount to ActiveUser fragment

B77Mills:
This flag is only set to true the first time a user has the verified flag set.  meaning this should only be true the first time a person clicks on the loginLink sent to their inbox.

Related PRs: 
 - https://github.com/parameter1/identity-x/pull/39
 - https://github.com/parameter1/randall-reilly-websites/pull/556

### DIFF
--- a/packages/marko-web-identity-x/api/fragments/active-user.js
+++ b/packages/marko-web-identity-x/api/fragments/active-user.js
@@ -6,6 +6,7 @@ fragment ActiveUserFragment on AppUser {
   id
   email
   verified
+  isFirstTimeVerifying
   givenName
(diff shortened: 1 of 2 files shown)


B77Mills:
[Add new event trigger & listener for the first time conversion of a user from non-verified to verified on loginLinkSent/Authenticate.](https://trello.com/c/bS1VKhAo/224-add-new-event-trigger-listener-for-the-first-time-conversion-of-a-user-from-non-verified-to-verified-on-loginlinksent-authentica)

solocommand (approved):

PR #741, opened 2023-07-11
Allow NAME sort for default + passed in sortField, `marko-web-search`

Shinsina:
NOTE: This does not updated the list of fields within `marko-web-search-sort-by` so as not to introduce unexpected behavior to existing uses of the component (tied exclusively to site searchs).

This change is in order for directories using the underlying `MarkoWebSearchConfig` to do so using `NAME` as a valid sort field.


Shinsina:
[Add sort option to directory layout](https://trello.com/c/sKtQOziN/287-add-sort-option-to-directory-layout)

brandonbk (approved):

PR #744, opened 2023-07-14
Prevent recaptcha badge css from being purged

B77Mills:
This will restore the badge being hidden.

Will allow for a dep upgrade to replace this: https://github.com/parameter1/randall-reilly-websites/pull/553

image

### DIFF
--- a/packages/marko-web-theme-monorail/scss/components/_site-newsletter-menu.scss
+++ b/packages/marko-web-theme-monorail/scss/components/_site-newsletter-menu.scss
@@ -24,9 +24,11 @@
   }
 }
 
+/*! purgecss start ignore */
 .grecaptcha-badge {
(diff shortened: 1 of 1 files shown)


B77Mills:
[reCaptcha badge should be hidden](https://trello.com/c/aTq9WXUb/251-recaptcha-badge-should-be-hidden)

solocommand:
@B77Mills Making this change without also adding the reCAPTCHA text to the user-submitted forms will bring us out of compliance. Ref https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed. If you're choosing to hide the badge on all sites, the user-facing text needs to be updated.

brandonbk (approved):

PR #757, opened 2023-08-02
Adjust download form display of region & postal code depending on country codes

B77Mills:
Only allow for the visibility of the regionCode & postalCode inputs when the country code is set to US, Canada or Mexico.  

Screen Shot 2023-08-02 at 11 31 16 AM
Screen Shot 2023-08-02 at 11 31 09 AM

### DIFF
--- a/packages/marko-web-identity-x/browser/custom-column.vue
+++ b/packages/marko-web-identity-x/browser/custom-column.vue
@@ -85,7 +85,7 @@
     />
     <template v-else-if="fieldKey === 'regionCode'">
       <region-code
-        v-if="showContryCodeFields"
+        v-if="showCountryCodeFields"
         v-model="user.regionCode"
         :country-code="user.countryCode"
         :class-name="className"
@@ -95,7 +95,7 @@
     </template>
     <template v-else-if="fieldKey === 'postalCode'">
       <postal-code
-        v-if="showContryCodeFields"
+        v-if="showCountryCodeFields"
         v-model="user.postalCode"
         :class-name="className"
         :required="required"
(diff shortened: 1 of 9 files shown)


B77Mills:
[White paper form - additional fields](https://trello.com/c/HCvZdouF/330-white-paper-form-additional-fields)

packages/marko-web-identity-x/browser/custom-column.vue
```diff
@@ -200,11 +215,25 @@ export default {
...
+    className: {
+      type: String,
+      default: '',
+    },
   },
   /**
    *
    */
   computed: {
+    countryCode() {
```
solocommand, line 227: You should do the same for postal code, if you are making the field show/hide based on the selected country.

packages/marko-web-identity-x/browser/custom-column.vue
```diff
@@ -200,11 +215,25 @@ export default {
...
+      return this.user.countryCode;
+    },
+    /**
+     *
+     */
+    showContryCodeFields() {
```
solocommand: typo
```suggestion
    showCountryCodeFields() {
```

packages/marko-web-identity-x/browser/form/fields/address-extra.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <form-group class-name="col-md-4">
+  <form-group :class-name="className">
```
solocommand, line 2: Can you verify that this prop is already present in all of the other components you added it to in `custom-column.vue`? If not, it should be added.

solocommand (changes requested):

packages/marko-web-identity-x/browser/form/fields/address-extra.vue
B77Mills: @solocommand  this has been added to the field components that where missing it.

solocommand (approved):

PR #113, opened 2021-06-28
Add ability to pass customMetadata to google structured data ld_json tag

B77Mills:
In the case of FCP they are adding  articleSection: primarySection, creators: [author.name] & keywords: [taxonomy.name].

**FCP Company:**
Screen Shot 2021-06-28 at 11 30 07 AM
**FCP Article**
Screen Shot 2021-06-28 at 11 30 26 AM
**Standard Article**
Screen Shot 2021-06-28 at 11 35 02 AM

PR #777, opened 2023-08-29
Remove email field from Company structured data

Shinsina:
This was originally added via: https://github.com/parameter1/base-cms/pull/113

**NOTE:** If we would like to add an aliased version of the `publicEmail` field called `email` (likely to prevent any backwards breaking behavior that could come with removing `email`) then that can certainly be done, initially just wanted to remove the actual trouble spot before making any additional alterations.

### DIFF
--- a/packages/marko-web/components/page/metadata/content.marko
+++ b/packages/marko-web/components/page/metadata/content.marko
@@ -73,9 +73,6 @@ fragment ContentPageMetadataFragment on Content {
     mobile
     publicEmail
   }
-  ... on ContentCompany {
-    email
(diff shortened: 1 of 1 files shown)


Shinsina:
[Remove private email from structured data](https://trello.com/c/C7sLV7nE/233-remove-private-email-from-structured-data)

brandonbk (changes requested):
@Shinsina - instead of removing email, lets change it to publicEmail and then update the ld-json to use this field instead (when present).

brandonbk (approved):

PR #784, opened 2023-09-12
Ensure expired content is respected on `WebsiteScheduledContent` queries

Shinsina:
PRODUCTION:
Screenshot from 2023-09-12 15-45-31
Screenshot from 2023-09-12 15-45-38

DEVELOPMENT:
Screenshot from 2023-09-12 15-45-16

It's possible we may want to add this as some sort of input and default it to this new behavior so as to not be backwards breaking.

### DIFF
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -1456,10 +1456,15 @@ module.exports = {
         ],
       };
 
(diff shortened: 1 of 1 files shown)


Shinsina:
[Expired content incorrectly showing up on topics pages](https://trello.com/c/ONq85oO8/175-expired-content-incorrectly-showing-up-on-topics-pages)

PR #807, opened 2023-09-28
Add sorting by activeCustomFieldsIds on profile page

B77Mills:
This will add the sorting based on the order set in the array within the given sites config when the activeCustomFieldsIds value is present.


B77Mills:
[Update DrB + LabPulse profile fields - 9/28/23](https://trello.com/c/jZVq20C1/410-update-drb-labpulse-profile-fields-9-28-23)

brandonbk (approved):

PR #814, opened 2023-10-18
Redirect nativeX story paths to non-trailing /

B77Mills:


packages/marko-web-native-x/middleware/with-story.js
```diff
@@ -13,12 +14,17 @@ module.exports = ({
   template,
   queryFragment = defaultFragment,
 } = {}) => asyncRoute(async (req, res) => {
+  const { query } = req;
+  if (req.path.substr(-1) === '/' && req.path.length > 1) {
+    return res.redirect(301, applyQueryParams({ path: req.path.slice(0, -1), query }));
+  }
+
```
solocommand: Give this a try, really no reason to copy/paste that old utility everywhere. Also just return nothing if you're trying to break execution early. Assuming you added `return` to the marko render to fix the eslint issue?

```suggestion
  // Remove trailing slashes
  const { query, path, url } = req;
  if (path.substr(-1) === '/' && path.length > 1) {
    const query = url.slice(path.length);
    const safe = path.slice(0, -1).replace(/\/+/g, '/');
    res.redirect(301, `${safe}${query}`);
    return;
  } 
```

solocommand (changes requested):

solocommand (approved):

PR #828, opened 2023-11-15
Add content-page fragment support for factory call

B77Mills:
currently we are mostly using this on content routes this will now allow for

queryFragment: queryFragment,

OR

queryFragment: queryFragment.factory({ useLinkInjectedBody: true })

Allowing it to not be backwards breaking as well as allowing you to configure it on a per condition based when needed.

Example site update: https://github.com/parameter1/watt-global-media-websites/pull/282

### DIFF
--- a/packages/marko-web-theme-monorail/graphql/fragments/content-page.js
+++ b/packages/marko-web-theme-monorail/graphql/fragments/content-page.js
@@ -7,34 +7,23 @@ const factory = ({ useLinkInjectedBody = false } = {}) => {
     id
     name
     teaser(input: { useFallback: false, maxLength: null })
+    labels
     body(input: { useLinkInjectedBody: ${useLinkInjectedBody} })
     published
-    labels
+    updated
     siteContext {
       path
       canonicalUrl
-      url
     }
     company {
       id
       name
-      siteContext {
-        path
-      }
+      canonicalPath
       enableRmi
-      primaryImage {
-        id
-        src(input: { options: { auto: "format,compress" } })
-        alt
-        caption
-        credit
-        isLogo
-      }
     }
     primarySection {
       id
       name
-      fullName
       alias
       canonicalPath
       hierarchy {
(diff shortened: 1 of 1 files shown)


packages/marko-web-theme-monorail/graphql/fragments/content-page.js
```diff
@@ -1,165 +1,208 @@
...
 
-module.exports = gql`
-fragment ContentPageFragment on Content {
-  id
-  name
-  teaser(input: { useFallback: false, maxLength: null })
-  labels
-  body
-  published
-  updated
```
solocommand, line 11: Updated date should not be removed.

packages/marko-web-theme-monorail/graphql/fragments/content-page.js
```diff
@@ -1,165 +1,208 @@
...
-    canonicalPath
-    hierarchy {
+    teaser(input: { useFallback: false, maxLength: null })
+    body(input: { useLinkInjectedBody: ${useLinkInjectedBody} })
+    published
+    labels
+    siteContext {
+      path
+      canonicalUrl
+      url
```
solocommand: Why is the non-canonical URL being included here?

packages/marko-web-theme-monorail/graphql/fragments/content-page.js
```diff
@@ -1,165 +1,208 @@
...
-    starts
-  }
-  ... on SidebarEnabledInterface {
-    sidebars: sidebarStubs {
-      name
-      body
-      label
+    primaryImage {
+      id
+      src(input: { options: { auto: "format,compress" } })
```
solocommand: Are we intentionally removing the `useCropRectangle` flag  and fields here? Pretty sure this will break cropped images...

solocommand: Same for crop dimensions and primary image display

packages/marko-web-theme-monorail/graphql/fragments/content-page.js
```diff
@@ -1,165 +1,208 @@
...
+    ... on ContentPodcast {
+      transcript
+    }
+    ... on ContentNews {
+      source
+      byline
+    }
+    ... on ContentEvent {
+      endDate
+      startDate
```
solocommand: What change will this have on events?

packages/marko-web-theme-monorail/graphql/fragments/content-page.js
```diff
@@ -1,165 +1,208 @@
...
+    }
+    ... on ContentArticle {
```
solocommand: What about other content? Restore the SidebarEnabledInterface

packages/marko-web-theme-monorail/graphql/fragments/content-page.js
```diff
@@ -1,165 +1,208 @@
...
+            name
+            type
+            siteContext {
+              path
+            }
+          }
+        }
+      }
+    }
+    images(input: { pagination: { limit: 100 }, sort: { order: values } }) {
```
solocommand: This will break bizbash

solocommand (changes requested):
@B77Mills there are a ton of backward-breaking changes here. In short, you cannot remove any fields from the default content-page fragment, and we seem to be adding a lot of fields that are specific to WATT, that don't make sense to be used across every site.

The default content page fragment is intended to be used as a baseline, and is _supposed_ to be customized by consuming sites, if different fields are needed.

This should be an entirely new fragment -- and I'd argue it should be reviewed heavily before rolled out to any other sites.

solocommand (approved):
Looks good with those changes.

PR #743, opened 2023-07-13
Add primaryCategory to sitemap url resolver

Shinsina:
This was preventing URLs to resolve in the same fashion they do within fields returned within `Content.siteContext` (`path`, `url`, `canonicalUrl`) and causing the sitemap URLs for Products and Companies (as defined within https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/definitions/platform/content/types/company.js#L19 and https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/definitions/platform/content/types/product.js#L9 are the only types that are allowed to utilize this field) to resolve incorrectly.

BEFORE (Product resolution):
Screenshot from 2023-07-13 14-13-11

AFTER (Product resolution):
Screenshot from 2023-07-13 14-13-26

Showing Article resolution is unimpacted:
Screenshot from 2023-07-13 14-21-14

Query used within `indm_multi` tenant using the IEN site context via `x-site-id` header.
`query {
  contentSitemapUrls(input: {since: "1689000000000", contentTypes:[Product]}) {
    loc
  }
}`

PR #830, opened 2023-11-29
Include alias on sitemap url query projection.

Shinsina:
Likewise to https://github.com/parameter1/base-cms/pull/743. This was preventing URLs to resolve in the same fashion they do within the fields returned within `Content.siteContext` (`path`,`url`,`canonicalUrl`) and causing the sitemap URLs for Pages (as defined within https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/definitions/platform/content/types/page.js#L16 and is the only type allowed to use this field) to resolve incorrectly.

BEFORE:
Screenshot from 2023-11-29 13-20-01

AFTER:
Screenshot from 2023-11-29 13-20-19

### DIFF
--- a/services/graphql-server/src/graphql/resolvers/platform/content.js
+++ b/services/graphql-server/src/graphql/resolvers/platform/content.js
@@ -1097,6 +1097,7 @@ module.exports = {
 
       const projection = {
         type: 1,
+        'mutations.Website.alias': 1,
         'mutations.Website.slug': 1,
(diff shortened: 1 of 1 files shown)


Shinsina:
[MNET - URLs not in site map](https://trello.com/c/MBVjyCKE/457-mnet-urls-not-in-site-map)

solocommand (approved):

PR #833, opened 2023-12-15
Move global monorail user templates out of site org repos

B77Mills:
Move out of org repos into shared theme-monorail repo

### DIFF
--- a/packages/marko-web-theme-monorail/templates/user/authenticate.marko
+++ b/packages/marko-web-theme-monorail/templates/user/authenticate.marko
@@ -1,0 +1,16 @@
+$ const { config } = out.global;
+
+$ const type = "authenticate";
+$ const title = "Authenticate";
+$ const description = `Complete your ${config.siteName()} user profile`;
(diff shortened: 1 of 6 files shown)


Shinsina:
@B77Mills Is the intent to replace existing global package level implementations across the board? This would break something like Mundo's login page that requires/utilizes an i18n translation to properly translate the page to Spanish.

Ref: https://github.com/parameter1/pmmi-media-group-websites/blob/master/packages/global/templates/user/login.marko 
https://github.com/parameter1/pmmi-media-group-websites/pull/476

B77Mills:
@Shinsina  That is the intent.  In the case of sites like mundo or PMMI as an org they can continue to use their own version or we figure out how to configure it.  Orgs like RR can use all but the profile template they have overwritten & then just update the /templates/user/index.js file to import them accordingly.  They would also have to update the start server call the local folder routes too.   

We can discuss how to better configure for this, but this should cover the standard ones.

PR #839, opened 2023-12-27
Migrate error handling for `deleted` content to after redirect handling, `withContent` middleware

Shinsina:
Due to the content routes across all sites matching on 8 digit IDs, additional handling must be done here in order to verify that the 8 digits passed in are indeed a valid Base content ID for the given tenant to allow for additional handling elsewhere, by which deleted items can be handled via the redirectToFn function by which will then redirect based on the returned item when applicable otherwise these will error with a distinct "published content" qualifier in their error message.

In my testing this cannot simply be handled in the Express redirectHandler pass-through function outlined/passed through via these files: https://github.com/parameter1/base-cms/blob/master/packages/marko-web/start-server.js#L120 https://github.com/parameter1/base-cms/blob/master/packages/marko-web/express/error-handlers.js#L55 https://github.com/parameter1/base-cms/blob/master/packages/marko-web/express/get-redirect.js#L36 as the content routing hard errors prior to these being reached when a content item is not found with the appropriate status due to this 8 digit ID match.

Screenshot from 2023-12-27 17-15-54

### DIFF
--- a/packages/marko-web/middleware/with-content.js
+++ b/packages/marko-web/middleware/with-content.js
@@ -17,7 +17,7 @@ module.exports = ({
   pathFn,
   formatResponse,
   sideloadDataFn,
-  contentIdStatusExceptions,
+  contentIdStatusExceptions = [],
 } = {}) => asyncRoute(async (req, res) => {
   const id = isFn(idResolver) ? await idResolver(req, res) : req.params.id;
   const { apollo, query } = req;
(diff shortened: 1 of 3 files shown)


Shinsina:
[Excessive amount of 404 Not Found pages](https://trello.com/c/55FR5oOU/412-excessive-amount-of-404-not-found-pages)

B77Mills (approved):

solocommand (approved):

PR #841, opened 2024-01-05
Properly support all allowed input fields `magazineScheduledContent` query

Shinsina:
This allows all the fields defined here: https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/definitions/platform/content/index.js#L648 to be properly utilized via the respective resolver.

PR #843, opened 2024-01-08
Allow all input fields in `magazine-scheduled-content` block-loader

Shinsina:
Missed in https://github.com/parameter1/base-cms/pull/841 this now updates the block-loader utilized by the sites to accept the same input as documented here: https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/definitions/platform/content/index.js#L648


Shinsina:
[Digital Edition changes](https://trello.com/c/JiWInAEQ/438-digital-edition-changes)

B77Mills (approved):

PR #869, opened 2024-02-14
Support for alias based search routing

Shinsina:
When `rootAlias` is passed in and set to something aside from `search` MarkoWebSearch, will now operate via an "alias based" approach, by which something like `directory/engineering-design-construction-services?contentTypes=COMPANY`

Will result in the following: 
image

If this behavior is opted into use of `marko-web-search-website-sections-filter` will need to be replaced with `marko-web-search-website-sections-by-alias-filter` and can otherwise be used identically.

This needs to be use _in conjunction_ with the `assignedToWebsiteSectionIds` configuration option to load the initial section list into `marko-web-search-website-sections-by-alias-filter`.


Shinsina:
[Directory UI/UX updates](https://trello.com/c/P7msvUGy/460-directory-ui-ux-updates)

brandonbk (approved):

PR #873, opened 2024-02-16
Add IdentityX Entity prop to IdX submit events

solocommand:
- Adds entity to all server route responses
- Adds entity into event payloads via mixin
- Sets `idxEntity` on P1E props
- Moves login link click event to P1E library
- Removes invalid authenticate dispatch from omeda rapid identify event

### DIFF
--- a/packages/marko-web-theme-monorail/browser/index.js
+++ b/packages/marko-web-theme-monorail/browser/index.js
@@ -28,24 +28,6 @@ const setP1EventsIdentity = ({ p1events, brandKey, encryptedId }) => {
   p1events('setIdentity', `omeda.${brandKey}.customer*${encryptedId}~encrypted`);
 };
 
-const dispatchP1EAuthenticate = (args) => {
-  const {
-    loginSource: actionSource,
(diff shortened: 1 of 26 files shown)


packages/marko-web-theme-monorail/browser/index.js
```diff
@@ -28,24 +28,6 @@ const setP1EventsIdentity = ({ p1events, brandKey, encryptedId }) => {
   p1events('setIdentity', `omeda.${brandKey}.customer*${encryptedId}~encrypted`);
 };
 
-const dispatchP1EAuthenticate = (args) => {
-  const {
-    loginSource: actionSource,
```
B77Mills, line 33: @solocommand does the loginSource: actionSource setting need to be accounted for when moving this up into the list of conversion events?

solocommand: No, after review `source`, `loginSource` and `actionSource` are all set to the same value by the event emitter mixin, and `actionSource` is the preferred field for consistency. Good catch though!

PR #874, opened 2024-02-20
Fix EntityID generation, form access cookie

solocommand:
- Adds a default value to the `generateEntityId` function added in #873 
- Sets default value for IdX form access cookie on example website
- Enables comments on content body on example website


solocommand:
[Problems with Commenting on IEN](https://trello.com/c/LhTRV1ZV/485-problems-with-commenting-on-ien)

PR #846, opened 2024-01-11
Allow exclusion of content ids from most popular block

Shinsina:
Screenshot from 2024-01-11 16-32-15

PR #875, opened 2024-02-20
Allow inclusion of specific types only, when `contentTypes` is specified, Most Popular blocks, Monorail

Shinsina:
Ex: https://most-popular-content.base.parameter1.com/retrieve?tenant=bobit&realm=63a54060da01648223f94c58&types=article

REF: https://github.com/parameter1/base-cms/pull/846


Shinsina:
[Exclude "news" content type from Most Popular section on PoliceMag.com](https://trello.com/c/ziQ5gLfi/192-exclude-news-content-type-from-most-popular-section-on-policemagcom)

brandonbk (changes requested):
@Shinsina - In the interest of being consistent with other blocks, change this to be an optional input parameter of includeContentTypes.  That would also make it more flexible, in case there's a need for multiple kinds of most popular queries on the same site (Most popular products, etc)

brandonbk (approved):

PR #880, opened 2024-02-26
Support `mobileNumber` submission via profile form

Shinsina:
**REQUIRES: https://github.com/parameter1/identity-x/pull/50**


solocommand (changes requested):
The phone numbers should not share a field -- if the user wants to make both phone number and mobile number visible and required, they should be able to.

You'll also need to update the logic that pushes data to IdentityX when prepopulating the user data https://github.com/parameter1/base-cms/tree/master/packages/marko-web-omeda-identity-x/omeda-data

solocommand (dismissed):

PR #882, opened 2024-02-27
`newsletterEmailSchedules` query

Shinsina:
image

This returns the schedules sorted by deployment date from oldest to newest for the range provided.

### DIFF
--- a/services/graphql-server/src/graphql/definitions/email/schedule.js
+++ b/services/graphql-server/src/graphql/definitions/email/schedule.js
@@ -11,6 +11,7 @@ extend type Query {
     model: "email.Schedule",
     using: { contentId: "content.$id" },
   )
+  emailSchedulesForDateRange(input: EmailScheduleByDayQueryInput!): [EmailSchedule]!
 }
 
 extend type Mutation {
@@ -86,6 +87,12 @@ input ContentEmailSchedulesQueryInput {
 input EmailScheduleSortInput {
   field: EmailScheduleSortField = id
   order: SortOrder = desc
+}
+
+input EmailScheduleByDayQueryInput {
+  newsletterId: ObjectID!
+  startingStartOf: Date!
+  beforeEndOf: Date!
 }
 
 input QuickCreateEmailSchedulesMutationInput {
--- a/services/graphql-server/src/graphql/definitions/email/schedule.js
+++ b/services/graphql-server/src/graphql/definitions/email/schedule.js
@@ -93,6 +93,7 @@ input EmailScheduleByDayQueryInput {
   newsletterId: ObjectID!
   startingStartOf: Date!
   beforeEndOf: Date!
+  timezone: String # tz database format, e.g. America/Chicago
 }
 
 input QuickCreateEmailSchedulesMutationInput {
--- a/services/graphql-server/src/graphql/definitions/email/schedule.js
+++ b/services/graphql-server/src/graphql/definitions/email/schedule.js
@@ -11,7 +11,8 @@ extend type Query {
     model: "email.Schedule",
     using: { contentId: "content.$id" },
   )
-  emailSchedulesForDateRange(input: EmailScheduleByDayQueryInput!): [EmailSchedule]!
+  "Query for newsletter schedules by newlsetter product id and specified date range"
+  newsletterEmailSchedules(input: NewsletterEmailSchedulesQueryInput!): [EmailSchedule]!
 }
 
 extend type Mutation {
@@ -89,11 +90,14 @@ input EmailScheduleSortInput {
   order: SortOrder = desc
 }
 
-input EmailScheduleByDayQueryInput {
+"Input for retrieiving newsletter schedules for a specified time range"
+input NewsletterEmailSchedulesQueryInput {
+  "The newsletter product id"
   newsletterId: ObjectID!
-  startingStartOf: Date!
-  beforeEndOf: Date!
-  timezone: String # tz database format, e.g. America/Chicago
+  "The millisecond precision UNIX timestamp to start looking for schedules at (inclusive)"
+  after: Date!
+  "The millsecond precision UNIX timestamp to stop looking for schedules at (inclusive"
+  before: Date!
 }
 
 input QuickCreateEmailSchedulesMutationInput {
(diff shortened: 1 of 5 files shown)


Shinsina:
[GraphQL API improvements for Newsletters](https://trello.com/c/w8trVVWU/419-graphql-api-improvements-for-newsletters)

services/graphql-server/src/graphql/resolvers/email/schedule.js
```diff
@@ -15,6 +17,57 @@ const clearSeconds = (date) => {
...
+
+      if (!contentIds.length) return [];
+      const output = schedules.reduce((array, schedule) => {
+        const { content: contentItem, ...rest } = schedule;
+        if (contentItem && contentItem.oid) {
+          array.push({ ...rest, content: contentItem.oid });
+        }
+        return array;
+      }, []);
+      return output;
```
solocommand: What are we doing here?

There should be no active schedules that do not have active content. The shape of the returned `EmailSchedule` should not be modified. If you _have_ to make the content reference only the id, you need to create and return a new type. But the better approach is to just return the id field when querying.

You should just be able to use `return basedb.find(...)` here.

services/graphql-server/src/graphql/resolvers/email/schedule.js
```diff
@@ -15,6 +17,57 @@ const clearSeconds = (date) => {
...
+        sort: scheduleSort,
+        projection: {
+          _id: 1,
+          product: 1,
+          'content.$id': 1,
+          section: 1,
+          deploymentDate: 1,
+          sequence: 1,
+          status: 1,
+        },
```
solocommand: The 4th argument to a GraphQL resolver is the [info](https://www.apollographql.com/docs/apollo-server/data/resolvers/#resolver-arguments) param, which includes metadata about the requested fields at query time.

You should build this value dynamically based on the fields the user requested. There's no need to include `section` or `product` (and incur the db query to materialize their data) if the user only requested the `content` and `id` fields).

You can build this data automatically using the [buildProjection](https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/utils/build-projection.js) utility:
```js
emailSchedulesForDateRange: async (root, vars, context, info) => {
  // ..
  const projection = buildProjection({ info, type: 'EmailSchedule' });
```

services/graphql-server/src/graphql/definitions/email/schedule.js
```diff
@@ -88,6 +89,13 @@ input EmailScheduleSortInput {
   order: SortOrder = desc
 }
 
+input EmailScheduleByDayQueryInput {
+  newsletterId: ObjectID!
+  startingStartOf: Date!
+  beforeEndOf: Date!
+  timezone: String # tz database format, e.g. America/Chicago
```
solocommand: Include GQL docs here indicating the purpose of these input fields, and expected format
```gql
"input description"
input inputname {
  "The newsletter product id"
  newsletterId: ObjectID!
  # ...
}
  ```

Do the same for the query above.

solocommand (changes requested):
A few changes.

IMO, none of the timezone stuff should be included here. Unless this logic is identical to the query logic used by the newsletter app (and is kept in sync), it's not consistent. Better to document the expected behavior than try and force it.

The date input expects a timestamp and it _should_ be treated like any other date. Whatever the user sends for an `after` timestamp is the lower bound, and `before` is the upper bound.

The `EmailSchedule.deploymentDate` field does not have a rendered date variant, so the value is going to be returned as a timestamp anyway, and the timezone is moot.

I'd also recommend:
- calling this `newsletterEmailSchedules` to match [contentEmailSchedules](https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/definitions/email/schedule.js#L10)
- calling the input `NewsletterEmailSchedulesQueryInput`
- changing your date range input fields to `after` and `before`

services/graphql-server/src/graphql/resolvers/email/schedule.js
Shinsina: Whether or not it should be the case, "in the real world" there are schedules that do not have active content, otherwise the `contentStatus` field would always be 1 and to that point there wouldn't really be a reason to have it, to that point it would appear that even when a content item is set to draft (status 2) the `contentStatus` on it's respective email schedule still indicates 1 (published) which is inherently not correct.

Here are some "real world examples":
Screenshot from 2024-03-15 11-58-59
Screenshot from 2024-03-15 11-59-10


Lastly an example of `contentStatus` functioning incorrectly:
https://manage.parameter1.dev/content/edit/article/22879537
Screenshot from 2024-03-15 12-02-52
Screenshot from 2024-03-15 12-02-59

To directly answer the question this is actually done to get the resolution to work correctly as this winds up using the following: https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/dataloaders/utils/run-queries.js#L72 which if I had returned this result "raw" without errors due to the values being passed in as `ids` there being objects over numerical values. What was done here was simply to correct the resolution as I believe it's attempting to use this: https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/definitions/email/schedule.js#L30 which seemingly requires it to be an id reference in order to perform something akin to `basedb.find(...)` inherently.

services/graphql-server/src/graphql/resolvers/email/schedule.js
Shinsina: Thanks for pointing this out, I had only seen `getProjection` within `newsletterScheduledContent` and wasn't sure how to get a valid projection for this. This seems to have worked as I needed it to here.

services/graphql-server/src/graphql/resolvers/email/schedule.js
solocommand: Ok, I think i followed all of that. You still should not need to modify the values.

The `EmailSchedule.content` field uses the `refOne` directive, which uses the `extractRefId` utility function that will handle the conversion to a raw id from the dbref format: https://github.com/parameter1/base-cms/blob/master/services/graphql-server/src/graphql/directives/ref-one.js#L34

That may use the dataloader/query aggregator under the hood, but you don't need to change the reference format in order for it to work. What result are you seeing that makes you think it won't work correctly?

Compare it to the `contentEmailSchedules` query, which returns this type. The content field works as expected there, so what's different here?

image

Shinsina: It looks like this might have been an artifact of however I was resolving this before, potentially due to something missing in my projection, I believe this to be resolved and I have removed this unnecessary rebuilding via: https://github.com/parameter1/base-cms/pull/882/commits/48bdeb75dcb080be870cd21f8c62d3015a7924b7

solocommand (approved):

PR #889, opened 2024-03-13
Use footer width and height for footer logo dimensions

Shinsina:


solocommand (approved):
@Shinsina Should these values fall back to whatever is currently in use? Or is the plan to update all the sites configs when upgrading deps?

Shinsina:
> @Shinsina Should these values fall back to whatever is currently in use? Or is the plan to update all the sites configs when upgrading deps?

@solocommand My intent would be to update all the site configs.

brandonbk (approved):

PR #895, opened 2024-03-28
Upgrade native-x lib to 1.1.0

zarathustra323:


zarathustra323:
[Send native web creative metrics to P1 events](https://trello.com/c/NOtEDQ9u/374-send-native-web-creative-metrics-to-p1-events)

PR #908, opened 2024-04-29
Add ability to filter websiteScheduledContent query by company

B77Mills:
Add the ability to pass set company: id with the websiteScheduledContent query input.  this will then filter all results by that given companies content.   For use in situations like displaying company content within blocks like spec comparison tool.   Also for use within the following pr https://github.com/parameter1/ac-business-media-websites/pull/568


solocommand (approved):
LGTM. Addition is covered by an existing index, so there should be no performance impact

PR #911, opened 2024-05-16
Add check on activeContext & activeContext.user before using.

B77Mills:
Add this to prevent `An error occurred: FormError: Cannot read property 'id' of null`

At the moment this would need to be udpate on SMG & ACBM
Screenshot 2024-05-16 at 8 55 47 AM

### DIFF
--- a/packages/marko-web-identity-x/service.js
+++ b/packages/marko-web-identity-x/service.js
@@ -417,8 +417,12 @@ class IdentityX {
    * @returns {?string} The entityId of the active user/identity, if present.
    */
   async generateEntityId({ appId, userId } = {}) {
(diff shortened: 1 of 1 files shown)


solocommand (approved):
This should be fine. The original intent was to avoid running the loadActiveContext query if the appId and userId were supplied (both because it was unnecessary and because it would modify the detected user/identity state of the service.)

But I don't think that will matter, since that identity state is only valid for the scope of the request.

solocommand:
@B77Mills although looking at your screenshot, how is it possible for there _not_ to be a user context at this point? When there is PII displayed on a form?

PR #914, opened 2024-05-20
Adjust setting of uid to prevent id of null error

B77Mills:
prevent pulling of id off of null object and getting the folllowing error An error occurred: Cannot read property 'id' of null


B77Mills:

PR #923, opened 2024-06-20
Add userRegistration to content in GTM dataLayer

B77Mills:
Add the userRegistration to the content object being passed to the gtm layer.  This will allow for variables to be set and triggers to know when a piece of content has gates applied.

Example is within EQW's GTM container 

Defines a GTM variable from dataLayer content.userRegistration.isCurrentlyRequired. with a default of false.  Then in Olytics tag conditionally append data prior to sending to olytics.

### DIFF
--- a/packages/marko-web-gtm/components/context/content.marko
+++ b/packages/marko-web-gtm/components/context/content.marko
@@ -47,6 +47,10 @@ fragment ContentContextFragment on Content {
       }
     }
   }
+  userRegistration {
+    isCurrentlyRequired
+    accessLevels
+  }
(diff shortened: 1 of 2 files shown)


B77Mills:
[All Brands: Creating a workaround for our personalizations (through Omeda) conflict with gated content (P1 and IdentityX)](https://trello.com/c/fJ3maTo0/384-all-brands-creating-a-workaround-for-our-personalizations-through-omeda-conflict-with-gated-content-p1-and-identityx)

brandonbk (approved):

PR #928, opened 2024-07-09
Return empty string to render instead of null or undefined when using block loader

B77Mills:
Prevents null from showing when the most popular block is not returned in the case like on content.nrwa.org


B77Mills:
[Changing our ad manager - new code for content portal](https://trello.com/c/0WQoAe5n/73-changing-our-ad-manager-new-code-for-content-portal)

brandonbk (approved):

PR #951, opened 2024-08-19
Adjust mindful ad preview route with template/layout concepts & update core version with company name

B77Mills:
This is now setup similar to the other template layout version of pages we uses.  This version will allow you to utiliz the standard global template with the three previews or you can override this template while utilize the orgs/sites version of nativeX block calls 

the global version now says ${Advertisers’s Name} ad preview for the title & set preview titles as such.  Preview #1, Preview #2 & Preview #3.  

This will hopefully remove any concern about naming of positions or allow the groups an easier way to override this route. 

Loaded on example website base-cms:
Screenshot 2024-08-20 at 9 23 31 AM

Loaded on IEN:
ad-preview

### DIFF
--- a/packages/marko-web-theme-monorail/components/layouts/ad-preview.marko
+++ b/packages/marko-web-theme-monorail/components/layouts/ad-preview.marko
@@ -1,0 +1,83 @@
+$ const { log } = console;
+$ const { get, getAsObject, getAsArray } = require('@parameter1/base-cms-object-path');
+$ const fetch = require('node-fetch');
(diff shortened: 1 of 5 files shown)


jwade1327 (approved):

jwade1327 (approved):

B77Mills:
[Add new global preview route for mindful native website ads](https://trello.com/c/Kca9i737/254-add-new-global-preview-route-for-mindful-native-website-ads)

PR #960, opened 2024-09-13
Use getAsObject for incoming attributes from input

Shinsina:
Resolves the error in which if attributes is not sent `Object.entries` errors due to `attributes` being undefined.


Shinsina:
[Deployment ID Error](https://trello.com/c/pj1PaJLw/441-deployment-id-error)

B77Mills (approved):

PR #972, opened 2024-10-01
Add simple contentAccess byPass same as contentDownload

B77Mills:
Add same check we do for contentDownload but for contentAccess.  meaning it will bypass the profile page if refering loginSource was from a contentAccess login form submission.

It in theory will be used with a profile form submission on the content landing route instead.


B77Mills:

brandonbk (approved):

PR #973, opened 2024-10-01
Make contentAccess submission mimic contentDownload submission functionality...Will now update profile info.

B77Mills:
This change is to make contentAccess now work similar or identical to [contentDownload](https://github.com/parameter1/base-cms/blob/master/packages/marko-web-identity-x/browser/download.vue#L260).  Meaning it would push to /profile await the response and get the update user and then send the content and user to the content access endpoint to generate the db insertion of access.  Allowing for the reporting aspect of this.  

@todo: in digging into this I discovered that with the introduction of the custom forms it allows for questions to be asked that are not configured to show on the profile page.  Not a huge deal, but in doing such those question do not pass as activeCustomeIds for the profile config and are not saved to the user profile.

### DIFF
--- a/packages/marko-web-identity-x/browser/access.vue
+++ b/packages/marko-web-identity-x/browser/access.vue
@@ -285,7 +285,7 @@ export default {
         }
       } catch (e) {
         this.error = e;
-        this.emit('download-errored', { message: e.message });
+        this.emit('access-errored', { message: e.message });
       } finally {
         this.isLoading = false;
       }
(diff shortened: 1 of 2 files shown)


B77Mills:

solocommand (commented):
Where is `updateProfileOnAccess` being implemented? I don't see it in the vue component anywhere.

brandonbk:
> Where is `updateProfileOnAccess` being implemented? I don't see it in the vue component anywhere.

@solocommand - good catch; this has now been removed.

brandonbk (approved):

PR #976, opened 2024-10-03
Unify IdentityX profile updates from access & download components

B77Mills:
This will be used mainly when there is a non logged in, but identified user being passed in that we dont want to be able to update the profile but rather pre pop the form and allow for them to easily submit data to the access route.

Better explination of what the heck this should now work like:

Condition(nothing to work with, no cookies, no user logged in):
Screenshot 2024-10-03 at 8 58 19 AM

This would still render the coponents with no user context and force them to login... setting the related cookies. 

SO say you you did that and now are logged in: you would get the following: 
Screenshot 2024-10-03 at 9 38 44 AM

In this case you can also hit log out and it will still prepop your stuff as it still uses your idx cookie id to look up the user. 

The cookie state can sometimes cause odd conditionss meaning if the incoming query parmas and cookie don't match I'm not sure at the moment who wins. :( 

But your example with oly_enc_id being present with no cookies would work.  You could also get someones Idx(Richeles) info if you knew how to do it, but you can subit as her, but not change her record in identityX:
http://www-acbm-fcp.dev.parameter1.com:9721/covid-19/whitepaper/22921299/brians-test-expired-company-brians-company-whitepaper?oly_enc_id=1084G1356867G5F

You can submit it with her data, but it will only submit to content access & not update her profile.
Screenshot 2024-10-03 at 10 29 08 AM

### DIFF
--- a/packages/marko-web-identity-x/browser/access.vue
+++ b/packages/marko-web-identity-x/browser/access.vue
@@ -65,8 +65,8 @@
       <login
         :additional-event-data="additionalEventData"
         :source="loginSource"
-        :endpoints="endpoints"
         :redirect="redirect"
+        :endpoints="endpoints"
         :app-context-id="appContextId"
         :consent-policy="consentPolicy"
         :consent-policy-enabled="consentPolicyEnabled"
@@ -242,7 +242,7 @@ export default {
         return this.content.siteContext.url;
       }
       if (this.content) {
-        return this.content.id;
+        return `/${this.content.id}`;
       }
       return '/';
     },
--- a/packages/marko-web-identity-x/browser/access.vue
+++ b/packages/marko-web-identity-x/browser/access.vue
@@ -196,6 +196,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    updateProfileOnSubmit: {
+      type: Boolean,
+      default: true,
+    },
   },
 
   /**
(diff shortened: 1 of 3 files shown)


B77Mills:

B77Mills:

packages/marko-web-identity-x/browser/access.vue
```diff
@@ -229,6 +234,26 @@ export default {
...
 
+    /**
+     *
+     */
+    redirect() {
+      if (this.content.siteContext && this.content.siteContext.url) {
+        return this.content.siteContext.url;
+      }
+      if (this.content) {
+        return this.content.id;
```
solocommand: I _think_ this should be prefixed with a slash. Can you confirm what the login/profile components do with the redirect prop? Based on the other two values though, this probably should include the leading slash.

packages/marko-web-identity-x/browser/download.vue
```diff
@@ -228,6 +233,26 @@ export default {
...
 
+    /**
+     *
+     */
+    redirect() {
+      if (this.content.siteContext && this.content.siteContext.url) {
+        return this.content.siteContext.url;
+      }
+      if (this.content) {
+        return this.content.id;
```
solocommand: Same here

solocommand (changes requested):
Looks good, but fix redirect leading slashes if necessary

packages/marko-web-identity-x/browser/access.vue
B77Mills: @solocommand Based on what im reading the login component takes this and uses it when set or generates its own off of the window.locations... ${pathname}${search}${hash} I was setting this more as an insurance.

solocommand (dismissed):

PR #983, opened 2024-10-10
Port newseltter signup modal from SMG to monorail to use within other sites.

B77Mills:
At the moment I tried to port it as closely as I could to what was already setup for [SMG](https://github.com/parameter1/science-medicine-group-websites/pull/502). That said I did make some adjustment to use a cookeValueToMatch prop being set within the newseltter.modal config to base its display logic off of vs being hard set.  This could and im sure will change.  At the moment this doesn't allow for initial page visit trigger either. 

THis is an example for WaterDaily: https://github.com/parameter1/encore360-websites/pull/34
This would be an example of a new site include it: https://github.com/parameter1/transpire-media-websites/pull/83
This would be how SMG would need to be updated: https://github.com/parameter1/science-medicine-group-websites/pull/611

 
Screenshot 2024-10-11 at 11 31 16 AM

Screenshot 2024-10-10 at 3 49 27 PM

Screenshot 2024-10-10 at 3 15 04 PM


B77Mills:
[Setup newsletter signup modal](https://trello.com/c/fPQnD0u7/77-setup-newsletter-signup-modal)

brandonbk (approved):

PR #987, opened 2024-10-15
Add fallback font for PT Sans

B77Mills:


B77Mills:
[Investigate CLS issues & optimizations](https://trello.com/c/SCO6aQ1C/85-investigate-cls-issues-optimizations)

PR #1000, opened 2024-10-29
Initial setup of mindful package

B77Mills:
Initial pass was started within the watt site org.  The intent is to utilize this shared package instead individual site versions. 

https://github.com/parameter1/watt-global-media-websites/pull/415

should be able to delete the local packate and update line #9 in the [start-server](https://github.com/parameter1/watt-global-media-websites/pull/415/files#diff-acc0ae15f68026388b2f7d0a0eda601ab0f2cfd8514f19d0899da527cfd6ccb8R9) file to 
`const mindful = require('@parameter1/base-cms-mindful/marko-web/middleware');`


packages/mindful/marko-web/service.js
```diff
@@ -0,0 +1,24 @@
+const convertAdvertisingPostToNativeStory = require(('./utils/convert-advertising-post-to-native-story'));
+
+class MindfulMarkoWebService {
+  /**
+ * @param {object} params
+ * @param {import("./api-client.js").MindfulApiClient} params.client
```
zarathustra323: Should point to `../api-client.js` instead of `./api-client.js`

packages/mindful/marko-web/utils/convert-advertising-post-to-native-story.js
```diff
@@ -0,0 +1,51 @@
...
+    ...rest
+  } = advertisingPost || {};
+  const { label } = statusEdge && statusEdge.node ? statusEdge.node : {};
+  const isPublished = label === 'Published' && (new Date() >= new Date(publishedDay));
+  if (id && (isPublished || preview)) {
+    const { _id: imageId, src } = featuredImageEdge.node || { src: { settings: {} } };
+    const primaryImage = {
+      id: imageId,
+      src: src.url,
+      // @todo Determine if these are the correct values for focal point
```
zarathustra323: Both `@todo` comments can be removed since they focal point info is correct

zarathustra323 (changes requested):

zarathustra323 (approved):

